PR #41, opened 2014-10-21
Add schemas for some astropy.modeling models

nden:
As I am reading through this and trying out pyasdf I tried creating schemas for `astropy.modeling` models commonly used for WCS transformations. So here they are.

I think only models which don't define an analytical inverse need the `inverse` definition in the schema so only the polynomials have it here. If this is acceptable I'll add the rest of the polynomials.


nden:
Close this in favor of #49.

PR #43, opened 2014-10-22
Extend YAML schema

mdboom:
To support specifying the flow_style of sequences and mappings and the
style of strings.

Collalory to spacetelescope/pyasdf#43


embray:
Only just now saw this, but I like!

PR #45, opened 2015-01-07
Encoding

mdboom:
This explains how I think encodings/compression should work in ASDF.

@perrygreenfield, @embray: Comments encouraged.


source/file_layout.rst
```diff
@@ -169,24 +169,39 @@ Each block begins with the following header:
...
-  used space for the block (not including the header), in bytes.
+  used space for the block on disk (not including the header), in
+  bytes.
+
+- ``memory_size`` (64-bit unsigned integer, big-endian): The size of
+  the block when decoded, in bytes.  If ``encoding`` is all zeros
+  (indicating no encoding), it **must** be equal to ``used_size``.  If
+  there is an encoding, this is the size of the decoded block data,
+  which **may** be different from the size of the encoded block in the
+  ASDF file.
```
embray: I'm going to nitpick over what to call this.  The fact that this involves decompressing "in memory" I think is an implementation detail (for example, you could decompress directly to disk or something).  Maybe `actual_size` (if only because it reference a They Might Be Giants song)?

embray:
I have mixed feelings about this and need to give it some more thought.

On the one hand, I understand the intention behind grouping encoding information with the binary blocks in a separate YAML section.  This makes sense when breaking the blocks out in "exploded" form.  On the other hand, I think we need to be clear about when this sort of encoding should or should not be used.  I think it should be used *only* for things like binary compression that affects what bytes are read out of the block.  I'm less comfortable with using this for encodings that are dependent on how those bytes or interpreted in the first place.  For example some of the array-binary encodings, like the quantization encoding.  This *requires* that any array read out of this block is interpreted as floats of some kind or another (and since the encoding scheme requires specifying 32- or 64-bit floats that means the dtype has to be specified twice; it also means that assigning an array of any other dtype to this data block is meaningless).  Worse still is the case where the data is scaled; something I think would be of interest to have in the primary tree.

So now we have a situation where the description of the bytes themselves, and the semantic interpretation of those bytes, is stored in separate places.  Of course, these are different levels of "semantic" interpretations too.  So maybe it's okay.  But where do we draw that line?

mdboom:
Sorry -- I merged this by accident -- I think @embray's feedback is worth considering first.  I'm going to reset master...

source/file_layout.rst
mdboom: Yeah -- I'm not crazy about the name either.  Just couldn't come up with a better name.  `actual_size` is fine.  `data_size` might be ok also.

mdboom:
So, in case it's not clear, my implicit goal is to support the kinds of things that are possible to do in FITS (with the non-standard convention in cfitsio for compression), but to do it in a way that's more extensible.  One of the things it does is it quantizes tiles of an image independently (each with their own scale and offset) on the assumption that the deviation within a tile will be smaller than across the whole image.

So, I share your concern about duplicating information about the array in both the tree and the block.  In some sense, we already have this problem, since if the datatype specified in the tree isn't appropriate for the bytes in the block things will not work.  It may be "safer" in some sense to store the datatype in the block and have the tree use offsets and strides in multiples of that datatype rather than bytes as it does now.  

However, for encodings, I'm not sure storing the decompression information with each view in the tree will work.  Say, for example, one is doing the lossy tile quantization from FITS, and there are multiple subviews on a large array.  If the compression information is stored in the tree, then you have to be able to decompress it from the information stored in each view alone, and each view would have to be independently decompressable without decompressing the whole.  That's a much more complex problem to solve.

I agree the need for both "array-aware" encodings and "byte stream" encodings is kind of messy.  It's, of course, impossible to do any sort of useful lossy compression without knowing the shape and datatype of the array and being able to slice it into tiles etc. so now we have the block being aware of such details where before it was just a "bag of bits".  And that's not great.

perrygreenfield:
I'm worried that we may be trying to solve too many problems at the beginning. Perhaps we should narrow the goals initially (e.g., not deal with views of compressed arrays in any efficient way, particularly if we don't wall off more complex solutions in the future). Looking it at it this way, is one or the other approach exclusive of the other? If we choose one, do we necessarily prevent an option to use the other?

Offhand, it seems that a library that must deal with only decompressing needed tiles for a view (or a requested slice) is going to be fairly complex, and I'd like to avoid dealing with that now. Can we start assuming compressed arrays will be completely decompressed before views are used, if that doesn't preclude support for more sophisticated access in the future?

mdboom:
My initial implementation just had a single field describing the kind of compression used on the block.  That works fine for things like lossless zlib compression, and is dead simple.  But it does preclude the sort of "stacking" of compression algorithms that the CFITSIO convention for FITS supports -- and my goal was to make sure that could be described.  So you do, at a minimum, need a way of stacking the algorithms.  It seemed at first that a fixed-length field would be enough to list the algorithms used (as they will always be small in number), but then you realize that most lossy compression algorithms will need some kinds of additional information to decompress.  That's how I ended up with the little YAML headerlet to the block, because that is the only way I can see to be open-ended and general.  In fairness, that could also be in binary form in the block header somehow, but given that it's of arbitrary length with values that would be different depending on the algorithm, YAML makes some sense.

>  Can we start assuming compressed arrays will be completely decompressed before views are used, if that doesn't preclude support for more sophisticated access in the future?

That's indeed what this does -- by storing the information about tile compression with the block and not with each view in the tree, it means you have to decompress the whole block to access the view.  I think while that may not be "the best one could do", it's reasonable and much easier to implement etc.  The point I was making was that if you move the information about the compression into the tree and stored with each view (rather than the block) to address @embray's concern about information about the array datatype being stored in two places, you are basically forced to also implement compression algorithms that can work on partial data (you'd probably end up doing something like "keyframes" in video data where you would rewind to the nearest marker of some kind and then start from there), all of which is a much harder problem to solve.  Or you just end up duplicating the information about the compression verbatim in every view, which is *more* of a duplication problem in my view than this PR as it now stands.

The only other way I can think of to address @embray's concern is to change how views work.  Currently, all views point directly to a block.  If views instead pointed to other views in the tree, then the one "master" view of everything would point to the block and it could store information about compression.  That's closer to how Numpy array views work so it has some precedence as a reasonable approach.

mdboom:
Actually, now that I think of it, I think I have a way to get rid of the array-specific information in the block header, which I this addresses most of @embray's concerns.  There's no way to get around that in the implementation of a *writer*, but I think there's a way to keep it out of the *spec*.  Stay tuned...

perrygreenfield:
> 
> The only other way I can think of to address @embray's concern is to change how views work. Currently, all views point directly to a block. If views instead pointed to other views in the tree, then the one "master" view of everything would point to the block and it could store information about compression. That's closer to how Numpy array views work so it has some precedence as a reasonable approach.

That’s what I was wondering. If done that way, the view doesn’t have to duplicate the information as it is already layered on top of the one portal to the information (if I understand correctly). But I’ll wait to see what you are going to propose.

mdboom:
So, I've gone down a few rabbit holes on this and ended up back at the beginning :(

In attempt (A) I tried to change ASDF so that views (subarrays) are references to other arrays in the tree, where the bottom array of the stack then points directly to a block.  This bottom array would then become the place where any encoding or compression is described.  This turns out to really complicate things.  What if the view is included in the tree, but its base is not?  What if (and this really happens), the views immediate base isn't in the tree, but it's secondary base is?  (In the following diagrams, views that the user put in the tree have an asterisk):

```
A* -> B -> data_block
A* -> B -> C* -> data_block
```

You have to then start implicitly inserting arrays into the tree just so that they can refer to a block.  Which then gets one in mind of having a special list in the tree just to describe each block.  And once you've done that, you might as well do what's done here which is to just put that description with each block.  :)

The problem with this approach is compounded by the fact that reconstituting Numpy views from disk is very difficult.  While you can create a view on the underlying data buffer quite easily (using the `ndarray` constructor), you can't create a view while declaring another view as its direct base.  You *can* do it by converting the strides back into slice indices and then taking a slice of the base, but this only works if the dtype of the view is the same as its base, so it doesn't work for tables, and in general is more limited than what current ASDF can do.

In attempt (B), I tried to take the knowledge of arrays out of the encodings.  I'm not sure how one can implement things like tile-rearranging without knowing something about the shape.  I tried to be general where possible, i.e. tiling requires knowledge of the datatype itemsize, but not the actual datatype.  But quantizing does require knowing the underlying floating point type.  I don't really see a way around that.

So, I'm sort of back where I started, but I'm (a) open to other ideas and (b) more convinced that this approach may be the right one.

perrygreenfield:
I assume you mean in the example views defined in Python, but where someone is only specifically asking for the views A (in the first case), and A and C (in the second case) to be in the ASDF file. My assumption is that yes, implicitly you have to get all the intermediate ones in the file too. The first case would be a bit odd for a file I would think, in that someone wants a view of a larger buffer, but isn't interested in the whole buffer, but still wants it in the file. I don't understand why that would be desired (but I may be completely misunderstanding your comment). For the second, if someone had defined it that way in Python and somehow didn't want B in the file, I would expect the onus is on them to redefine A so that it doesn't depend on B. So is the issue here how transparent it should be to go from Python to ASDF, or the issue of organizing ASDF trees themselves?

As for reconstituting a  view from another view that is the base of an array, does that correspond to the above examples (or is that something like A*->B*->C -> data_block)? In other words, I implicitly have assumed that any data buffer better have a base object that is general enough for all views (even if that requires it to be a byte array). Again, I may be (and almost certainly am) completely confused :-)

mdboom:
> I assume you mean in the example views defined in Python, but where someone is only specifically asking for the views A (in the first case), and A and C (in the second case) to be in the ASDF file. 

That's correct.

> My assumption is that yes, implicitly you have to get all the intermediate ones in the file too. 

That seems like a big ask of users, since Numpy views can be created in so many ways -- often you aren't even aware you have a view.  The library could find all the views that aren't explicitly included in the tree and add them to a special place in the tree just for these "anonymous" views, or somehow skip over them by redefining them.  In either case, it provides a lot of implementation complexity and bookkeeping that isn't there now.

> The first case would be a bit odd for a file I would think, in that someone wants a view of a larger buffer, but isn't interested in the whole buffer, but still wants it in the file.

This is useful if the larger buffer exists in another file, i.e. in "exploded" form.  Another example might be a number of subarrays pointing to different parts of a larger buffer, totally the whole, but not having an array pointing to the complete buffer.  That's useful if you have a data cube that you want to represent as an array of 2d planes (something one of the jwst_lib.models does).

> So is the issue here how transparent it should be to go from Python to ASDF, or the issue of organizing ASDF trees themselves?

It certainly straddles implementation and design :wink:  In the current design, it's very easily to automatically convert views in Python to views in ASDF because the heirarchy is flat and guaranteed to have everything we need.  Moving to a views-are-based-on-other-views approach, the user will either have to be more explict about how the views should be in the file, or the system needs to be more careful about how it detects cases that can't be saved directly.

About reconstitution, if you had something like:

```
A -> B -> C -> data_block
```

saved in the file, Numpy doesn't really provide a way to create B based on C based only on strides and offset -- you can only B based on `data_block`.  There's no good reason it couldn't -- just not part of Numpy's current abilities.  You can create it correctly, but the dependence of B on C is lost, and only relates to `data_block`.  So you couldn't roundtrip an ASDF file without "flattening" the tree of views.  But all of that seems to be an artifact of Numpy, and if we really want views-on-views in the file format at the end of the day, we should just go about fixing Numpy.

> In other words, I implicitly have assumed that any data buffer better have a base object that is general enough for all views (even if that requires it to be a byte array)

That's correct in the current design.  The data block really is just a "blob of bits" and the views can look at it however they want.  If views could point to other views, Numpy places restrictions on only "valid" slices, so we'd have implementation difficult there, particularly with tables.  Numpy places no restrictions on viewing the underlying data buffer.

perrygreenfield:
So I guess one of the issues is whether to fold in numpy restrictions into this, and whether views of views have any merit in the first place. I was a little confused about your description, but I think you made it clear that although you can create numpy arrays based on nested views, the arrays have no internal indication of nesting so any such ASDF relationships are lost within numpy (and as such, even if numpy arrays were created by the user in a nested way, there isn't any way to save from the arrays that creation relationships since numpy doesn't retain it).

But I'm trying to tie this back to your rabbit hole explorations. For example:

A* -> B -> data_block
A* -> B -> C* -> data_block

Originally, I had assumed that the above relationships were those in numpy, since the asterisk refers only to items in the tree (B isn't being referred to in the tree). But if B isn't coming from numpy, where is it coming from? (If B was created in numpy, it isn't relevant to A any longer since A knows nothing of B).

So I'm still a bit murky on the context of this example (not that you aren't right about the way it has to work in ASDF).

mdboom:
> I was a little confused about your description, but I think you made it clear that although you can create numpy arrays based on nested views, the arrays have no internal indication of nesting so any such ASDF relationships are lost within numpy

Sorry -- I wasn't clear.  In numpy, one can create views of other arrays through slicing (and arbitrarily nest those views) even though it's a single data buffer underneath.  But you can't create views of views through offset and stride -- those can only be created on top of the underlying buffer, so the nested heirarchy is "flattened".  So there is is a way to introspect and *save* nested views, but there is no way to *restore* that nesting from disk, unless you express everything in terms of slicing rather than striding can do.  Unfortunately, slicing is more limited because Numpy won't let you do anything type-unsafe with them.

> Originally, I had assumed that the above relationships were those in numpy, since the asterisk refers only to items in the tree (B isn't being referred to in the tree). But if B isn't coming from numpy, where is it coming from? (If B was created in numpy, it isn't relevant to A any longer since A knows nothing of B).

Your original assumption is correct.  Those relationships are coming from Numpy.  Numpy *does* have the context about the whole chain from `A -> B -> C` originally.  If `B` isn't in the tree, the ASDF file doesn't have that context, so we either have to re-express `A` on top of `C` (in which case roundtripping won't be exact, but it probably doesn't matter to the user if they didn't put B in the tree) or include `B` in some non-user part of the tree (which just seems messy).  But in any event, not because it *can't*, but because it *doesn't*, we can't rebuild the `A -> B -> C` chain because Numpy doesn't provide an interface for that.  All we can do is `A -> data_block`, `B -> data_block` and `C -> data_block`.

In summary: as you say, it comes down to whether expressing nested rather than flat relationships are useful in ASDF.  I think they probably are useful -- it's a question of whether they are worth the complexity tradeoff.  Nested views introduce some necessary complexity (how you deal with missing links in the chain, how you garbage collect them, deal with cyclical references etc.) that don't have much to do with Numpy.  And they also introduce some problems that are specific to Numpy that we should ignore-and-then-fix-Numpy if there's compelling reasons to have nested heirarchies in ASDF.

But at the end of the day, whether you do that or not, to support encodings/compression, you are going to have exactly one "chunk of information" per block describing how to decompress/decode/reorder it.  Whether that info is in the main YAML tree or at the beginning of the block, it can really only exist one place.  It might as well just go with the block -- unlike the stuff currently in the tree which humans could edit without too much fear of breaking, if they know what they are doing, changing the description of the block encoding without also re-encoding the block doesn't make much sense.  So I'm starting to feel that the issue of nested views is somewhat orthogonal to the issue of encodings anyway.

perrygreenfield:
Yes, I tend to agree. And I'm biased towards what is easier to implement than theoretically possible (particularly if the latter isn't precluded later). It's much better to get something that works the way we say it does out much sooner (and remember, the complexity will be judged by how hard it is to do in C, not Python as far as wider adoption goes).

mdboom:
@embray, @perrygreenfield: Stepping away from this problem for a few days, I think I have a new way forward, but want to get your thoughts first before devoting too much time on an implementation.

What if we pared down the "block encoding" part of this to just the most simple lossless compression -- only things like zlib that work on streams of binary data and don't have to have a deep understanding of its structure.  That's all pretty simple and obvious as to how it should work.  In this case we can just say what the compression algorithm is in a field in the block header -- no need for extra YAML there.

Then, for the higher-level stuff lossy stuff, which does need a deep understanding of the array structure, we reuse the `transforms` approach that we're using for WCS -- but it wouldn't be part of the WCS, it would just be a raw `transform` object hanging off of any array.  And *probably* this transformation would happen automatically on loading of the file, as opposed to a WCS.  There's no reason we couldn't define a transform that "shuffles" an array into tiles.  Defining a transform that converts int+scale+offset to floats would be a little outside of the box, but not impossible either.  (The tricky bit there is that the "input" data type to the transform is different from the "output" data type, so both would have to be defined somehow).  The code for these transforms obviously wouldn't have to live in `astropy.modeling`, which would still be for more "mathematical" transformations rather than these "computer sciencey" ones.  I like the idea because we already have the bones of a pretty sophisticated transformation infrastructure, so let's reuse that, rather than invent a new one over here for compression.  Of course, if we allow the user to put arbitrary transformations on any array, it's possible they'll put a rotation in there, or something, which would require resampling -- and resampling of course opens up a whole kettle of worms.

This gets tricky with views, but the same problems that currently exist with our WCS transforms would exist here, too.  In short, I think *all* transforms probably need to be defined in terms of the whole array as well, if they are going to be reusable at all.  That is, given a full array, `chip`, and a view into it, `subarray`, you probably want to do something like:

```
chip: !core/ndarray
   wcs: &my_wcs # Some big long WCS description

subarray: !core/ndarray
   wcs: @my_wcs
```

The only way to make this aliasing work (where both the full and the view refer to the same WCS) is if the WCS input coordinates are in terms of the full array.  We have the same requirement for compression-like transformations.

If this makes any sense at all, I think my way forward would be to (a) pare this PR down to the basics for stream-based compression (at which point we can say we've addressed the bare minimum for compression support), (b) deal with the rest as a separate step.

perrygreenfield:
I�ll have to give it some more thought, but it sure seems logical and well matched to the processes and needs. I like it.

Perry

PR #49, opened 2015-02-03
add polynomial schemas

nden:
This PR adds schemas for polynomials implemented in modeling.
After playing with these for a while I'm in favor of not dealing with multiple polynomials as implemented in modeling and keep it as general as possible. If this approach is accepted I'll create schemas for the rest of the polynomials. I will also submit a PR implementing these in pyasdf shortly.
@mdboom

### DIFF
--- a/source/schemas/transform.rst
+++ b/source/schemas/transform.rst
@@ -41,6 +41,15 @@ Arithmetic operations
    stsci.edu/asdf/0.1.0/transform/divide.rst
    stsci.edu/asdf/0.1.0/transform/power.rst
 
+Simple Transforms
+-----------------
+
+.. toctree::
+   :maxdepth: 1
+
+   stsci.edu/asdf/0.1.0/transform/shift.rst
+   stsci.edu/asdf/0.1.0/transform/scale.rst
+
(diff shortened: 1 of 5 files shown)


source/schemas/transform.rst
```diff
@@ -50,3 +50,13 @@ Projections
...
+
+Polynomials
+-----------
+
+.. toctree::
+   :maxdepth: 1
+
+   stsci.edu/asdf/0.1.0/transform/polynomial.rst
+   stsci.edu/asdf/0.1.0/transform/shift.rst
+   stsci.edu/asdf/0.1.0/transform/scale.rst
```
mdboom: Do `shift` and `scale` belong in the Polynomials section?  I see how they sort of are simple polynomials, but I'm not entirely sure one would look for them there.  Maybe under a new heading "simple transforms" or something?

mdboom:
This looks good.  Nice and general, like you say.

source/schemas/transform.rst
nden: I moved them to a section "Simple Transforms".

PR #53, opened 2015-02-25
Add domain on transforms

mdboom:
This adds a `domain` key to all transforms.

@nden: If this makes sense to you, I'll go ahead and add an implementation in pyasdf (which would just stuff the information in `meta.domain` until `astropy.modeling` grows support for this).

### DIFF
--- a/schemas/stsci.edu/asdf/0.1.0/transform/domain.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/transform/domain.yaml
@@ -21,12 +21,18 @@ examples:
 
 properties:
   lower:
-    description: The lower value of the domain.
+    description: >
+      The lower value of the domain.  If not provided, the
+      domain has no lower limit.
     type: number
+    default: -inf
 
   upper:
-    description: The upper value of the domain.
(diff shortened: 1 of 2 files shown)


schemas/stsci.edu/asdf/0.1.0/transform/domain.yaml
```diff
@@ -0,0 +1,39 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/0.1.0/transform/domain"
+tag: "tag:stsci.edu:asdf/0.1.0/transform/add"
```
nden: Did you mean "domain" here, not "add"?

nden:
I think this is good for now. Thanks.

schemas/stsci.edu/asdf/0.1.0/transform/domain.yaml
mdboom: Oops.  Yes.

PR #55, opened 2015-02-25
Simple compression specification

mdboom:
This is a replacement for the complex encoding scheme in #47 with just the bare minimum and obvious approach to losslessly compress a block.

The plan is to use the transform conventions to support lossy compression that needs to know about the array at some later date.

@embray @perrygreenfield


embray:
We never did get a chance to sit down and talk about this.  Sorry about that--I just got distracted with other things.  I hope to have a look at this soon.

mdboom:
Yeah -- this is an attempt to boil it all down to its (hopefully) uncontroversial core to just have something...

migueldvb:
Is it possible to use other data compression methods such as lossy altorithms for some kind of data?

mdboom:
@migueldvb: The long term plan includes lossy compression, but finding a good approach to that is proving difficult.  It's complicated by the fact that any lossy algorithm needs access to the array data types and structure which blurs the lines between the binary blocks and the tree description of those blocks.  This second attempt is *just* to deal with the lossless side of things and postpone the lossy stuff for later.  There is a good chance the lossy stuff will be based on the existing `transform` set of schemas.  See the discussion in #47 for more info.

mdboom:
@embray: Any thoughts?  I think this core should be uncontroversial -- and the lossy stuff will be designed/implemented later based on transforms.

migueldvb:
@mdboom Thanks for the clarification about lossy compression. I think that the data compression with zlib looks good.

embray:
Sorry for the lag on my part, but :+1: to this.  

As for lossy compression schemes I think that should be handled entirely in the tree, since on a certain level that becomes just as much part of the science.  I'm not 100% sure how the specifics should look though. Definitely something we'll have to look at more.

PR #56, opened 2015-02-26
Add schema for 3D rotations.

nden:


mdboom:
This looks good.  Eventually, we'll probably want to include a really explicit definition of the math, something like:

http://mathworld.wolfram.com/EulerAngles.html

But that shouldn't hold up the PR.  Nothing else, other than the polynomial stuff, really has that yet, anyway...

PR #57, opened 2015-02-26
add direction to rotate3d

nden:


nden:
Jut a note that this isn't ready to merge yet.

nden:
@mdboom I added the standard euler rotation sequences here. Let me know if you prefer them in a separate schema.

mdboom:
I don't think it needs to be a separate schema -- but maybe add a short description of what each value means?

nden:
Done. I think not all transform schemas are showing in the built documentation, not sure why. For example the projections are missing.

schemas/stsci.edu/asdf/0.1.0/transform/rotate3d.yaml
```diff
@@ -6,7 +6,16 @@ tag: "tag:stsci.edu:asdf/0.1.0/transform/rotate3d"
...
+  Euler angle rotation around 3 axes.
+
+  `direction` : Sequence of rotation axes: one of `zxz`, `zyz`, `yzy`, `yxy`, `xyx`, `xzx`
+                or `native2celestial`, `celestial2native`
+
+  `phi`, `theta`, `psi`: - Euler angles in degrees.
+                If `direction` is one of `native2celestial` or `celestial2native`,
+                `phi`, `theta` are the longitude and latitude of the native pole in
+                the celestial system and `psi` is the longitude of the celestial pole in
+                the native system.
```
mdboom: These descriptions should probably go with the properties themselves (below).

mdboom:
Just tried building your branch, and all of the transformation schemas (including the projections) do seem to be included.

nden:
I don't see them (or at least expected to see more) here

http://asdf-standard.readthedocs.org/en/latest/schemas/transform.html#projections

mdboom:
Ah -- that seems to be something readthedocs specific.  I'll see if the logs say anything.

mdboom:
I've fixed readthedocs.

nden:
I've nothing to add here.

PR #58, opened 2015-03-02
Markdown

mdboom:
It seemed silly to support restructuredtext in the schema descriptions, since that's a fairly Python-specific tool.  This changes things so it use the more universally-known markdown syntax.

Adds a dependency (just to build the standard docs) on `mistune` to parse the markdown.

### DIFF
--- a/md2rst.py
+++ b/md2rst.py
@@ -1,0 +1,195 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+"""
+Implements a markdown to reST converter using the mistune markdown
+parser.
+"""
+
+import re
+import textwrap
+
+import mistune
+
+
+class BlockLexer(mistune.BlockLexer):
+    # Adds math support
+
+    def __init__(self, *args, **kwargs):
+        super(BlockLexer, self).__init__(*args, **kwargs)
+        self.enable_math()
+
+    def enable_math(self):
+        self.rules.block_math = re.compile(r'^\$\$(.*?)\$\$', re.DOTALL)
+        self.default_rules = ['block_math'] + mistune.BlockLexer.default_rules
+
+    def parse_block_math(self, m):
+        self.tokens.append({
+            'type': 'block_math',
+            'text': m.group(1)
+        })
+
+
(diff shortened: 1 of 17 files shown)


embray:
I'm not crazy about markdown and its lack of standardness either.  But I agree overall with the intent behind this change.

embray:
Also are we saying use of markdown in schema descriptions should be part of the standard?  Or is this just a local convention we happen to be using...

mdboom:
markdown isn't so bad if you stick to the core features...

I think I plan to say markdown is the standard when writing schemas -- that way tools to manipulate them can rely on that being the case.

One of the use cases I have in mind (which I had meant to mention here, but didn't) was a tool that would allow you to explore an ASDF document and pull up pop-up help about a particular element from its schema.  Assuming such a tool is written as a web app, markdown is a lot easier to work with than reST (as the IPython guys can attest to, I'm sure).

embray:
That much I believe.  I just wish it were more "standard", and so have some misgivings about adding it to the format.  Not that I would advocate reST either.  It benefits in terms of "standards" by having only one implementation (that I know of).  But agree it's a bigger pain to work with, especially for the web.

For lack of a better suggestion I don't want to say I'm against it.  Maybe I'd feel better if we specified a strict subset of "markdown" that should be supported.

md2rst.py
```diff
@@ -0,0 +1,195 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+"""
+Implements a markdown to reST converter using the mistune markdown
+parser.
+"""
```
embray: Also, this is pretty nice!

migueldvb:
I think that markdown is a good choice for schema descriptions. There is an effort to standardize markdown by a group of people that are working on a reference syntax specification:
http://spec.commonmark.org/
I think there is another independent effort by the IETF.

md2rst.py
mdboom: Yeah, mistune is a pretty clean design.  I become aware of it because apparently IPython is now using it...

embray: Ah, cool.  At first I wondered if you were using pandoc, but this seems simpler.

PR #60, opened 2015-03-02
Minor documentation improvements

mdboom:


source/schemas/index.rst
```diff
@@ -17,3 +17,13 @@ must support the "core" module, but the other modules are optional.
...
+
+The following graph shows the dependencies between modules:
+
+.. digraph:: modules
+
+   "fits" -> "core"
+   "unit" -> "core"
+   "transform" -> "core"
+   "wcs" -> "transform"
+   "wcs" -> "unit"
```
embray: :+1: I wonder if these relationships shouldn't be made explicit in the schemas?  In a way I guess they already are implicitly, but maybe a field in the schema documentation for this is not a bad idea (as a separate property so that it could in principle be used in the future for introspection).  Maybe getting ahead of myself though...

mdboom: That's not a bad idea.  They are explicitly documented in the .rst files that pull in the schema documentation, but having it the YAML would let tools programmatically get at the info.

PR #67, opened 2015-03-10
Fix #64: Describe updatibility

### LINKED ISSUE #64: modify and append blocks to file
It is not completely clear if the layout allows to extend a dataset in a single self-contained file by modifying blocks and appending new blocks without rewriting the entire file. We are interested in an application that requires appending new data by updating the tree which could have empty space before the start of the blocks, and appending data to unused space in blocks and new binary blocks at the end. Cuold this write operations be done at end-of-file in single-file mode without rewriting the entire file?

mdboom:
Cc: @migueldvb


migueldvb:
:+1: thanks for the description.

PR #70, opened 2015-03-11
Add % for hexadecimal character

migueldvb:


mdboom:
I think this should be `0x` as a prefix in both cases, actually...

migueldvb:
That makes sense. I have changed the prefix.

mdboom:
The test failure is unrelated: see #71

PR #75, opened 2015-03-20
Split the affine matrix into linear transformation matrix and a translation

nden:
Addresses spacetelescope/pyasdf#72. 
This helps when using it to represent a PC matrix transformation in the context of WCS.


nden:
@mdboom Any objections to this change?

mdboom:
Is is still an affine transformation if it doesn't have the translation components?  Maybe this should be a new type...

embray:
I don't see the point of having a separate type for an affine transformation without translation since you could just default the translation to zero.

mdboom:
Oh -- and I misread the patch anyway.  I thought this was just removing the translation rather than separating it out into another key.

I guess I'm fine with this in that case.

embray:
Oh I see--yes, this makes sense.

PR #77, opened 2015-03-31
add SelectorMask schema

nden:
@mdboom and the schema for `SelectorMask`.


nden:
@mdboom Assuming selector schemas will live in asdf-standard, this is ready.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - type: object
+    properties:
+      mask:
+        description: |
+          An array with the shape of the detector/observation.
+          Pixel values are of type int or str and represent
+          region labels as in the RegionSelector transform.
+          Pixels which are not within any region have value 0 or " ".
+
+        $ref: ./selector_mask
```
mdboom: Did you intend mask to be another transform, or an array?  If the latter, I'm not sure why you have the selector_type here, rather then just saying `$ref: ../core/ndarray` and not having `selector_mask.yaml` below.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
```diff
@@ -0,0 +1,55 @@
...
+          Names of inputs.
+        type: array
+        items:
+          type: string
+      outputs:
+        description: |
+          Names of outputs.
+        type: array
+        items:
+          type: string
```
mdboom: What are the names of inputs and outputs about?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
```diff
@@ -0,0 +1,55 @@
...
+        items:
+          type:
+            - integer
+            - string
+      transforms:
+        description: |
+          A transform for each region. The order should match the order of labels.
+        type: array
+        items:
+          $ref: transform
```
mdboom: Maybe a mapping of `(label, transform)` pairs instead?  Easier to stay in sync that way, and the order doesn't matter in that case.

mdboom:
An example would be helpful here -- I suspect my questions are just from not quite understanding what you want this to end up looking like.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: I want mask to be a selector_mask model, so that an `inverse` can be assigned to it if needed.
Does this make sense?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: These are , for example, `inputs=('x', 'y')`, ` outputs=('ra', 'dec', 'lambda')` but are basically a way to get `n_inputs` and `n_outputs` in the model (which may be different). This is how the base class in modeling is set up but if @embray has better ideas, I'm listening.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: All right.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
mdboom: Why can't you just assign an inverse to the `regions_selector` (which would be another `regions_selector`)?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
mdboom: I see.  Because this is so general, it doesn't have those names like most of the transforms do.  But are these input and output names overridable on any transform?  In which case, we should probably define inputs/outputs in the base class (`transform.yaml`) and handle them in the base class in the Python code as well.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: I can and I am but there are cases when the transforms have an analytical inverse so there's no need to assign an inverse to the `RegionsSelector`. All one needs to do is assign an inverse to the `SelectorMask`.

mdboom: I see.  But as it's defined now, one has to do:

```
transform: !regions_selector
   mask:
      mask: [...array...]
      inverse:
          mask: [...array...]
```

which seems confusing to me.

Maybe have `mask` and `inverse_mask`, both of which are arrays?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: Yeah, it turns out they are settable. I'll add them to `transform.yaml` as part of this PR.

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: I think the inverse model rarely will be an instance of Selector Model, so most likely there won't be  another array as mask. For example, polygon regions on the detector turn into footprints on the sky. I haven't worked out the details of this yet, but perhaps the sphere package can be used for the inverse. Any ideas?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: Actually looking closer at this `inputs` and `outputs` don't work correctly in modeling. They are a settable property but there are no checks if the right number is passed.  So I am not going to add them after all. This schema works because transforms have `additionalProperties` set to true.

nden:
Also I just added an example and turned labels and transforms into a mapping.

So the types for these go in gwcs. Is a certain structure expected?
Similarly to pyasdf I put them in  gwcs/tags/. Is this correct?

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
mdboom: I guess I'm lost about how these two transform types related to one another, and the example doesn't help me understand.  Why the extra level of indirection i.e. `mask: mask: array`?  Where would the inverse go in the example and what would it look like?

mdboom:
Yes.  That structure makes sense -- not required, but we might as well be consistent with pyasdf.  I just added an example of how to extend pyasdf with new types here:  https://github.com/spacetelescope/pyasdf/pull/85

schemas/stsci.edu/asdf/0.1.0/transform/regions_selector.yaml
nden: @mdboom Back to this issue. Are you questioning the names or the models? I suspect it's the names and yes, this looks a bit repetitive :). The models for these schemas are here:

https://github.com/spacetelescope/gwcs/blob/master/gwcs/selector.py

A use case for `SelectorMask` is one in which it is initialized by vertices of polygons on the detector. The inverse is a list of spherical polygons on the sky. However I'm just now starting to play with the sphere package, I can't include an example yet. So I'm proposing the following changes:

Change the name of the parameter from `RegionsSelector.mask` to `RegionsSelector.selector_mask`.  So your example above would look like this:

```
transform: !regions_selector
   selector_mask:
      mask: [...array...]
      inverse:
          selector_mask:
              mask: [...array...]
```
 In addition I think it'd be best if `SelectorMask` keeps the `vertices` parameter for easier computation of spherical polygons, so the new signature may be
 `SelectorMask(mask=None, vertices=None, geometry=None)`. I have one more use case which I meed to support and it is computing the mask using {label: vertices} and {label:transform} . So obviously the interface will change in the future and so will the schema. 
But a possible schema may look like:
```
transform: !regions_selector
   selector_mask:
      vertices: [{label: vertices_array]
      geometry: cartesian
      inverse:
          selector_mask:
              vertices: [{label: spherical_polygon_vertices}]
              geometry: spherical
```

Or are you questioning the models in which case ignore this.
Obviously this schema is driven by the implementation and use cases and those are not quite set yet.

nden:
@mdboom I've updated this PR to match the code in spacetelescope/gwcs#18. Could you advise on the error here? Should I turn `dmapper` into an OrderedDict or remove it from the example?

mdboom:
The failure is a YAML syntax error.

I think since you have multi-line strings in your examples, as most do, you need to change:

```
    - A LabelMapperArray - array indices are mapped to labels.
      mapper = LabelMapperArray(np.array([[1, 0, 2],
                                          [1, 0, 2],
                                          [1, 0, 2]])
```

to

```
    - |
        A LabelMapperArray - array indices are mapped to labels.
        mapper = LabelMapperArray(np.array([[1, 0, 2],
                                          [1, 0, 2],
                                          [1, 0, 2]])
```

(and there are a few other instances of this).

nden:
Thanks! With that I was able to write out one of the more complex WCSs with regions to file!

nden:
@mdboom This PR was updated with the new versioning scheme.

schemas/stsci.edu/asdf/transform/label_mapper-0.1.0.yaml
```diff
@@ -0,0 +1,138 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/label_mapper-0.1.0"
+tag: "tag:stsci.edu:asdf/transform/label_mapper-0.1.0"
+title: >
+  Represents a _LabelMapper.
```
mdboom: The Python class `_LabelMapper` is just an implementation detail.  Maybe say something like:  "Represents a mapping from a coordinate value to a label."

schemas/stsci.edu/asdf/transform/label_mapper-0.1.0.yaml
```diff
@@ -0,0 +1,138 @@
...
+description: |
+  A label mapper instance maps locations to a label.
+  It is used together with RegionsSelector. The LabelMapper
```
mdboom: Here you use Python class names, but should probably use ASDF tag names.  You can refer to other tags in ASDF using markdown, e.g.:

```
It is used together with [regions_selector](ref:regions_selector)
```

schemas/stsci.edu/asdf/transform/label_mapper-0.1.0.yaml
```diff
@@ -0,0 +1,138 @@
...
+  returns the label corresponding to given inputs. The RegionsSelector
+  returns the transform corresponding to this label. This maps uniquely inputs
+  (e.g. pixels on a detector) to transforms.
+
+examples:
+  -
+    - |
+        A LabelMapperArray - array indices are mapped to labels.
```
mdboom: I don't know if it makes sense to include Python examples here -- the ASDF spec should probably be generic.  I'd just say: "Map array indices to labels" and not have the Python code snippet.

schemas/stsci.edu/asdf/transform/label_mapper-0.1.0.yaml
```diff
@@ -0,0 +1,138 @@
...
+            n_inputs: 2
+
+allOf:
+  - $ref: "transform-0.1.0"
+  - type: object
+    properties:
+      mapper:
+        description: |
+          An array with the shape of the detector/observation.
+          Pixel values are of type int or str and represent
```
mdboom: Should use JSON schema terms here, i.e. "Pixel values are of type integer or string and represent..."

nden:
@mdboom Thanks fo rthe comments. I've changed the wording and the examples. This goes together with spacetelescope/gwcs#19. Let me know if you have any comments there. I think tests there won't pass until this is merged.

nden:
@mdboom I forget what the status of this is. Is it going to make 1.0?

mdboom:
It can make 1.0 if it's ready...

nden:
The extension defined here

https://github.com/spacetelescope/gwcs/pull/19/files#diff-c6a62b84f2e9a4c32f72c50471e55d37

gives an error (the end of the traceback is below). The schemas live in asdf-standard but the readers/writers are in gwcs. I can't figure out why it's getting a wrong path. Help?
Changing `url_mapping` to the one below did not help either:

```
SCHEMA_PATH = os.path.abspath(
    os.path.join(os.path.dirname(pyasdf.__file__), 'schemas'))

def url_mapping(self):
        return [('http://stsci.edu/schemas',
                 util.filepath_to_url(SCHEMA_PATH) +
                 '/{url_suffix}.yaml')]
```

Failing command:
```
f=AsdfFile(extensions=extension.GWCSExtension())

-
    284 def _load_schema(url):
--> 285     with generic_io.get_file(url) as fd:
    286         if isinstance(url, six.text_type) and url.endswith('json'):
    287             result = json.load(fd, object_pairs_hook=OrderedDict)

/user/dencheva/ssbx-rhel6/lib/python2.7/site-packages/pyasdf-0.0.dev682-py2.7.egg/pyasdf/generic_io.pyc in get_file(init, mode, uri)
   1165                 fd = atomicfile.atomic_open(realpath, realmode)
   1166             else:
-> 1167                 fd = open(realpath, realmode)
   1168             fd = fd.__enter__()
   1169             return RealFile(fd, mode, close=True, uri=uri)

IOError: [Errno 2] No such file or directory: u'/user/dencheva/ssbx-rhel6/lib/python2.7/site-packages/pyasdf-0.0.dev682-py2.7.egg/pyasdf/schemas/stsci.edu/asdfcore/asdf-0.1.0.yaml'
```

mdboom:
The actual issue was the extra slash in the `tag_mapping` in extension.  But since your extension doesn't actually need to override the mappings (since it's using schemas in the main ADSF standard), it's best to just inherit from `BuiltinExtension` rather than `AsdfExtension` and not define any mappers at all.  See https://github.com/nden/gwcs/pull/1

nden:
Unless someone else has comments, this is ready to merge.

PR #78, opened 2015-04-03
flow_style -> flowStyle for consistency with camelCase of other keywords

embray:
If any associated changes need to be made in pyasdf I'll do that too.

<s>Relatedly, I was reading the descriptions of this, and the `style` keyword, and thought of adding a note that implementations *may* check this on reading too.  It would be up to implementations decide how to alert about objects that are not given in the specified style, though the recommended default is to ignore it.</s>

<s>I don't know, what do you think?  I feel like in general this should be ignored on reading, but I think there could also be value in using this for checking a file to *strict* adherence to convention, and that implementations should at least know that they have that option.  I believe this is possible to do within pyyaml but I'm not certain.</s>

On second thought, scratch all that.  That has the same problem as trying to preserve comments, I guess.  This would have to hook schema validation deep into the YAML parser in a way that might be doable but probably isn't practical in the current model.

### DIFF
--- a/schemas/stsci.edu/yaml-schema/draft-01.yaml
+++ b/schemas/stsci.edu/yaml-schema/draft-01.yaml
@@ -31,7 +31,7 @@ allOf:
         items:
           type: string
 
-      flow_style:
+      flowStyle:
         description: |
           Specifies the default serialization style to use for an
           array or object.  YAML supports multiple styles for


embray:
Looks like pyasdf already has it as `flowStyle`, so the bug is just in the metaschema I guess?

migueldvb:
I think this needs to be changed only in a couple of places in `pyasdf`.

mdboom:
Yeah, I guess this makes sense, given the JSON Schema tends to use camelCase.  I'll merge this and then make a corresponding PR over in pyasdf.

embray:
@mdboom Though since this was already correct in pyasdf, does that mean there are no examples/tests using flowStyle that also validate the example schema against the metaschema?

mdboom:
There are examples that test the example schema against the metaschema, but they don't fail because extra keys are allowed in the JSON metaschema.  So `flowStyle` isn't rejected, even though it was defined as `flow_style`.  A test that checked that expected a failure when `flowStyle` is some incorrect value would have caught this, though.

mdboom:
On closer look, I think it's also possible that this was tested at one time, but was stopped at some point.

Initially, all shemas were validated against the metaschema when they were loaded.  This turned out to be very slow, so I changed it so all of the ASDF schemas are validated as part of the test suite, but once installed are assumed correct.  This works well, except the ad hoc testing schemas that were used for testing `style` and `flowStyle`, since they aren't "ASDF schemas" were never being validated (and none of the ASDF schemas use `flowStyle` yet).

embray:
Ah okay, I was wondering about that.  I agree validating the metaschema is a waste to do every time, but for testing purposes we should try to.

Hmm...maybe in the future it would be nice to add to the implementation a cache tracking which schemas have been validated so that they can at least be validated once locally, but then not subsequently.

mdboom:
There already is a (memory) cache when it loads a schema.  But an on-disk cache of what has been validated is an interesting idea (but on disk caches are always surprisingly troublesome)...

embray:
That they are--I figured we could leverage some of the work done in Astropy, but that still has many open questions in its own right :/

PR #81, opened 2015-04-29
Fix Blaze broken link

migueldvb:
Fix  broken link for Blaze docs.


mdboom:
Thanks for catching.  It looks like `http://blaze.pydata.org/` redirects to `en/latest`, so maybe use that, as it's even less likely to change in the future?

migueldvb:
That's a good point. I have changed the link to `http://blaze.pydata.org/`

PR #83, opened 2015-05-05
Add table support

mdboom:
As mentioned in astropy/astropy#3733

Cc: @astrofrog, @embray

### DIFF
--- a/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
@@ -23,8 +23,7 @@ properties:
 
   data:
     description: |
-      The array data for the column.  The array must have a scalar
-      datatype (that is, it may not be a structured array).
+      The array data for the column.
     allOf:
       - $ref: ndarray
 
--- a/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
@@ -14,7 +14,10 @@ description: |
 type: object
 properties:
   name:
-    description: The name of the column
+    description: |
+      The name of the column.  Each name in a
+      [table](http://stsci.edu/schemas/asdf/0.1.0/core/table) must be
+      unique.
     type: string
     pattern: "[A-Za-z_][A-Za-z0-9_]*"
 
--- a/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/core/column.yaml
@@ -1,0 +1,47 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/0.1.0/core/column"
+tag: "tag:stsci.edu:asdf/0.1.0/core/column"
+
+title: >
+  A column in a table.
+
+description: |
+  Each column contains a name and an array of data, and an optional description
+  and unit.
+
+type: object
+properties:
+  name:
+    description: The name of the column
+    type: string
+    pattern: "[A-Za-z_][A-Za-z0-9_]*"
+
+  data:
+    description: |
+      The array data for the column.  The array must have a scalar
+      datatype (that is, it may not be a structured array).
+    allOf:
+      - $ref: ndarray
+
+  description:
+    description: |
+      An optional description of the column.
+    type: string
+    default: ''
+
+  unit:
+    description:
+      An optional unit for the column.
+    allOf:
+      - $ref: ../unit/unit
+
+  meta:
+    description:
+      Additional free-form metadata about the column.
+    type: object
+    default: {}
+
+requiredProperties: [name, data]
+additionalProperties: false
--- a/schemas/stsci.edu/asdf/0.1.0/core/table.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/core/table.yaml
@@ -1,0 +1,108 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/0.1.0/core/table"
+tag: "tag:stsci.edu:asdf/0.1.0/core/table"
+
+title: >
+  A table.
+
+description: |
+  A table is represented as a list of columns, where each entry is a
+  [column](ref:http://stsci.edu/schemas/asdf/0.1.0/core/column)
+  object, containing the data and some additional information.
+
+  The data itself may be stored inline as text, or in binary in either
+  row- or column-major order by use of the `strides` property on the
+  individual column arrays.
+
+  Each column in the table must have the same first (slowest moving)
+  dimension.
+
+examples:
+  -
+    - A table stored in column-major order, with each column in a separate block
+    - |
+        !core/table
+          columns:
+          - !core/column
+            data: !core/ndarray
+              source: 0
+              datatype: float64
+              byteorder: little
+              shape: [3]
+            description: RA
+            meta: {foo: bar}
+            name: a
+            unit: !unit/unit deg
+          - !core/column
+            data: !core/ndarray
+              source: 1
+              datatype: float64
+              byteorder: little
+              shape: [3]
+            description: DEC
+            name: b
+          - !core/column
+            data: !core/ndarray
+              source: 2
+              datatype: [ascii, 1]
+              byteorder: big
+              shape: [3]
+            description: The target name
+            name: c
+
+  -
+    - A table stored in row-major order, all stored in the same block
+    - |
+        !core/table
+          columns:
+          - !core/column
+            data: !core/ndarray
+              source: 0
+              datatype: float64
+              byteorder: little
+              shape: [3]
+              strides: [13]
+            description: RA
+            meta: {foo: bar}
+            name: a
+            unit: !unit/unit deg
+          - !core/column
+            data: !core/ndarray
+              source: 0
+              datatype: float64
+              byteorder: little
+              shape: [3]
+              offset: 4
+              strides: [13]
+            description: DEC
+            name: b
+          - !core/column
+            data: !core/ndarray
+              source: 0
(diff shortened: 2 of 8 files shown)


schemas/stsci.edu/asdf/0.1.0/core/column.yaml
```diff
@@ -0,0 +1,47 @@
...
+description: |
+  Each column contains a name and an array of data, and an optional description
+  and unit.
+
+type: object
+properties:
+  name:
+    description: The name of the column
+    type: string
+    pattern: "[A-Za-z_][A-Za-z0-9_]*"
```
embray: Are the restrictions on column names in order to not conflict with possible programming language syntax and the like?  That makes sense, of course, but it still seems overly restrictive, though maybe fine for now as a practical concession.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+  data:
+    description: |
+      The array data for the column.  The array must have a scalar
+      datatype (that is, it may not be a structured array).
```
embray: Okay, but that still allows repeated/multi-dimensional values in columns, right?

embray:
Wow, that was impressively fast, and more or less exactly what I had in mind.  Looking good!

embray:
If we add this I guess maybe we'll have to update the paper as well?  Maybe not significantly, but just to mention it as one of the core types.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
mdboom: That's exactly the reason -- to make sure that the column names can be used as token identifiers in any reasonable programming language (note that it's even more restrictive than what Python 3 would support for example).  I think it's an important distinction -- one of the main flaws of VOTable IMHO is that it doesn't do this, and you have to jump through all kinds of messy hoops to make the columns addressable on the Python side.  I wouldn't be opposed to having both this name and some sort of "display name" which would be unrestricted, though, if it's necessary (I can particularly think of the usefulness of Greek math symbols, for example).

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
mdboom: I'm not sure what to do about this.  This just came from the restriction/design of `astropy.table.Table`.  But that might be letting the horse design the cart, so maybe ASDF shouldn't explicitly disallow it.  I don't know how difficult it would be to get `astropy.table.Table` to support recarray columns.

mdboom:
Good point about updating the paper.  I sort of plan to make a "prerelease" of the standard to correspond with the paper once the final version is submitted -- not because it represents any particular milestone, but more just to have something fixed to point to from the paper so people will know when the paper and the standard disagree that we are talking about different versions of the standard.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
astrofrog: If you do go with a normal name and a display name, please don't make it so the display name is optional but unique and the normal name is required but not unique <cough>VO</cough> :wink:

embray: Yeah I like the idea of adding a separate display name.  Display name could be optional.  Both names should be required to be unique!

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
embray: I'm not necessarily talking about general recarrays (i.e. structs).  Just columns where the values are, for example, 5x5 arrays.  That *is* supported by `Table` (and in fact was on the of the drivers for its creation).  

I don't think recarray columns make a whole lot of sense, but maybe someone could come up with a reason for it.  Fine with leaving it out for now.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
mdboom: Not that this should drive the ASDF design, but I don't think `astropy.table.Table` currently supports a display name, so we'd need to add support for that (or shove it in `meta`).

@taldcroft: Do you have any thoughts on adding a display name to Table columns?  I doing this would be most useful if `Columns` understood them and used them in the `repr` etc.  Though I could definitely see the downside to this in terms of confusion (having two, possibly very different, names for the same thing).   I think I'm personally -0 on the idea because of that, but I could be convinced otherwise.

astrofrog: I personally have never had the need for a display name. I can see some possible minor benefits, but I'd have to be convinced they are worth the infrastructural complexity.

taldcroft:
@mdboom - I also have never found a real use case for a column display name that is different from the actual name.  I guess one can imagine wanting to have a latex-formatted version of a column name, but given the code complexity it would introduce I would want to see some stronger drivers.

taldcroft:
Just out of curiosity, what would be the proper way in ASDF to identify two columns as the primary RA and Dec in a source list table?  We seem to have this discussion frequently in astropy regarding automatically identifying these in a table.  Is the idea to mark a table with a particular registered schema?  Does the VOTable concept of a UCD have any place here?

schemas/stsci.edu/asdf/0.1.0/core/table.yaml
```diff
@@ -0,0 +1,108 @@
...
+    - |
+        !core/table
+          columns:
+          - !core/column
+            data: !core/ndarray
+              source: 0
+              datatype: float64
+              byteorder: little
+              shape: [3]
+            description: RA
```
taldcroft: Not a big deal, but might be nice to have the example(s) be a bit more realistic by having column names `ra`, `dec`, `id` and descriptions `RA (J2000)`, `Dec (J2000)`, `Source identifier`.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
```diff
@@ -0,0 +1,49 @@
...
+    allOf:
+      - $ref: ndarray
+
+  description:
+    description: |
+      An optional description of the column.
+    type: string
+    default: ''
+
+  unit:
```
taldcroft: Do you want a `format` key as well?  This is maybe a philosophical question of whether a column should have an intrinsic property of how to best represent itself to the user.  I like this idea, but I'll admit that in practice I very rarely set the column format in my day to day work.

mdboom: I think that's fine as an optional feature.  VOTable has something sort of like that for example.  But I think that can easily be added later, too.  The trick will be to not choose a Python-specific format language -- maybe just use C's printf as a basis since that's about as widely supported as anything.  Or use some abstract thing like VOTable (where precision and sig. digits are separated out etc.)

mdboom:
> Just out of curiosity, what would be the proper way in ASDF to identify two columns as the primary RA and Dec in a source list table? We seem to have this discussion frequently in astropy regarding automatically identifying these in a table. Is the idea to mark a table with a particular registered schema? Does the VOTable concept of a UCD have any place here?

I have wondered how UCD might fit into all this.  What is the general consensus about UCD in the community?  Is it widely used?  I don't think I have enough domain knowledge to judge how useful or problematic it is.  I'd certainly prefer to reuse it if it makes sense, rather than reinvent the wheel.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
embray: Okay, for now we can assume "YAGNI".  I admit for most astronomy data it seems like it would be a rarity.  But since ASDF is aiming to be useful beyond astronomy the lack of international support for column names makes me sad.

mdboom: @embray: But the purpose of the column name is to be able to address it uniquely from some sort of software code, and for that purpose I think an "ASCII token" is still the best for practical reasons.

We can always add a display name later, along with UCD or something like it for more info.

embray: Sorry, my last comment was unclear.  I was referring specifically to having a more open-ended display name.

embray:
Do want to go ahead and merge this, or do you think it needs something else?  ISTM the UCD question can be resolved separately.

schemas/stsci.edu/asdf/0.1.0/core/column.yaml
embray: I like this idea too.  FITS tables have something similar as well (it's never been supported in pyfits, though I would like to).  Definitely agree it should be optional, but I also have no reservations about keeping representation hints in the metadata (ASDF schemas include such information as well).

schemas/stsci.edu/asdf/0.1.0/core/table.yaml
```diff
@@ -0,0 +1,108 @@
...
+      A list of columns in the table.
+    type: array
+    items:
+      $ref: column
+
+  meta:
+    description: |
+      Additional free-form metadata about the table.
+    type: object
+    default: {}
```
embray: So now that I play around with this, I'm thinking maybe that something else the table schema needs to allow is a `data` property.  That way we can write all the data for a table in one place as an inline array that the individual columns can reference.

Currently if you write a table with inline, each column's data is written inline with the column, but we lose the nice inline row-based table.  I think an argument could be made for both representations, but I think probably keeping the row-based output will be more common.  

One thing that isn't immediately obvious to me is how to tell the writer which way to write the table.  I fear an additional write option might be necessary for that :/

mdboom: Yes, I think this makes sense.

You're right: it adds another mode -- "inlining of columns" vs. "inlining of entire tables".  I think the latter is probably going to be more common and maybe should be the default behavior for the already existing "inline".  We could try to get away with only the latter until the former because obviously necessary.

mdboom:
Yes -- I think we can deal with UCD later.  (UCD's may make sense on arrays as well, so they aren't really a table-specific issue anyway).

PR #85, opened 2015-05-26
Custom validators

mdboom:
This adds some custom schema validators related to ndarrays.

### DIFF
--- a/schemas/stsci.edu/asdf/0.1.0/asdf-schema.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/asdf-schema.yaml
@@ -13,11 +13,11 @@ allOf:
     properties:
       max_ndim:
         description: |
-          Specifies that the matching ndarray is at most the given
-          number of dimensions.  If the array has fewer dimensions, it
-          should be logically treated as if it were "broadcast" to the
-          expected dimensions by adding 1's to the front of the shape
-          list.
+          Specifies that the corresponding ndarray is at most the
+          given number of dimensions.  If the array has fewer
+          dimensions, it should be logically treated as if it were
+          "broadcast" to the expected dimensions by adding 1's to the
+          front of the shape list.
         type: integer
         minimum: 0
 
--- a/schemas/stsci.edu/asdf/0.1.0/asdf-schema.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/asdf-schema.yaml
@@ -1,0 +1,46 @@
+%YAML 1.1
+---
+$schema: "http://json-schema.org/draft-04/schema"
+id: "http://stsci.edu/schemas/asdf/0.1.0/asdf-schema"
+title:
+  ASDF schema
+description: |
+  A metaschema extending YAML Schema and JSON Schema to add support
+  for some ASDF-specific checks, related to nd-arrays.
+allOf:
+  - $ref: "http://stsci.edu/schemas/yaml-schema/draft-01"
+  - type: object
+    properties:
+      max_ndim:
+        description: |
+          Specifies that the matching ndarray is at most the given
+          number of dimensions.  If the array has fewer dimensions, it
+          should be logically treated as if it were "broadcast" to the
+          expected dimensions by adding 1's to the front of the shape
+          list.
+        type: integer
+        minimum: 0
+
+      ndim:
+        description: |
+          Specifies that the matching ndarray is exactly the given
+          number of dimensions.
+        type: integer
+        minimum: 0
+
+      datatype:
+        description: |
+          Specifies the datatype of the ndarray.
+
+          By default, an array is considered "matching" if the array
+          can be cast to the given datatype without data loss.  For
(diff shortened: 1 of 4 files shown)


schemas/stsci.edu/asdf/0.1.0/asdf-schema.yaml
```diff
@@ -0,0 +1,46 @@
+%YAML 1.1
+---
+$schema: "http://json-schema.org/draft-04/schema"
+id: "http://stsci.edu/schemas/asdf/0.1.0/asdf-schema"
+title:
+  ASDF schema
+description: |
+  A metaschema extending YAML Schema and JSON Schema to add support
+  for some ASDF-specific checks, related to nd-arrays.
```
embray: Interesting; deep.

Reminds me I need to get back to making a YAML Schema standalone release.

I'm a little surprised by the need for this, but I think it makes sense.

mdboom: In `jwst_lib.models`, we want to be able to set up schema that says "This array must be two-dimensional, because I'm expecting a CCD image" etc.  We've actually had this feature for a long time over there, but now all of those underpinnings are being replaced with pyasdf, and I think it's general enough to include here.  But if it adds too much complexity, it could be a "private" feature of `jwst_lib.models`.

embray: I keep bouncing back and forth on this.  Can you explain in your own words why it has to be part of a metaschema, and can't just be part of the ndarray schema?

mdboom: This allows us to write schemas that validate the type or dimensions of an array.  Presently, all we can say is that it *is* an array.  And we can say what keys an array description should have.  But we need to extend the metaschema in order to add new validation types.  The use case for this is in the JWST pipeline, where we want to say a particular file contains a 2D image or a 3D ramp or 4D cube etc.  But this is already being used internally (in this PR) where you see it can declare that the data type of a mask is boolean.

mdboom: And to just elaborate a little bit more -- this is no different than what YAML schema does to add new validation keywords to JSON schema.  This is just adding some more -- but since these are specific to ASDF (specifically to how ndarrays work) and are not generic to YAML, I added them in a new "layer" on top of YAML schema.  So we have a heirarchy of `ASDF schema -> YAML schema -> JSON schema`.

embray: The layering of metaschemas I understood.

I guess what I'm confused by (and I didn't sleep well last night so I'm probably off-focus) is the need for special properties for this.  For example ndarray objects already have a shape property.  So is not possible for a schema to require something like:
```yaml
tag: ndarray
properties:
    shape:
        minItems: 2
        maxItems: 2
```

(or something to that effect; I forget the exact syntax).

That covers both `ndim` and `max_ndim`.  Seems like you could do something similar with datatype.

Anyways that said, having these special keywords is more convenient and I'm not against it.  I'm just trying to make sure I understand the thinking behind it.

mdboom: Doh.  Not sure why I didn't see that with `ndim` and `max_ndim`.

`datatype` requires more complex logic though -- since (at least for the JWST pipeline use case), we want to be able to say things like "at least 32 bit integers" rather than an exact match.  If we expressed datatypes numerically (not as strings) then that might be possible.  But there's still the corner case of float-to-int casting where it's a little more complicated than just the number of bits.

mdboom: So, I definitely get your point (why include a tool for something we can do already), but it sure makes things verbose as I try to convert things over on the jwst_lib.models end.  The exact incantation (equivalent to ndim) is:

```
allOf:
   - $ref: <...>
   - properties:
      shape:
         minItems: 2
         maxItems: 2
```

vs.

```
ndim: 2
```

(since in the latter case, the fact that it's an array is also conveniently implied).

embray: I agree--I still think the ASDF schema is nicer here.  I do worry about a slippery slope, where suddenly every schema author wants their own custom validators for their schema, and it becomes impossible for implementations to keep up.

But if we keep this at an "ASDF core" level (and in particular since this is for something as low-level and fundamental as array dimensions) I think it's fine.

mdboom: I think if third parties want to add validators for their custom types, they are still free to do so -- they will simply be ignored by our validator, but it doesn't make those custom types unusable by tools that don't understand them.

The pyasdf implementation of this actually makes it easy for third-party code to add validators to the system (as part of the type definition).  That only applies to pyasdf, of course.

In any event, I agree with you -- since this is so strongly tied to a core type, it's probably fine to include.

Issue #20, opened 2014-06-16
Support extended precision floating point numbers


mdboom:
Following Numpy's lead on this is going to lead to confusion, so it's been backed out.

But at some point, we should find a way to support the landscape of different floating point numbers -- probably as an "optional" feature, since not all programming environments will find it very convenient.

PR #87, opened 2015-06-01
Add support for a time type

mdboom:
This provides a way to represent time in ASDF.

@taldcroft: This is mostly based around what astropy does (with some simplifications to remove things that aren't really relevant to a file format).  I'd really appreciate your thoughts on this one (and there are some TODO questions below).

### DIFF
--- a/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
@@ -1,6 +1,6 @@
 %YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+$schema: "http://stsci.edu/schemas/asdf/0.1.0/asdf-schema"
 id: "http://stsci.edu/schemas/asdf/0.1.0/time/time"
 title: Represents an instance in time.
 description: |
@@ -25,8 +25,7 @@ description: |
   In either case, a single time tag may be used to represent an
   n-dimensional array of times, using either an `ndarray` tag or
   inline as (possibly nested) YAML lists.  If YAML lists, the same
-  format must be used for all time values [TODO: This is a restriction
-  of astropy: Do we need to retain it or should we try to fix that?]
+  format must be used for all time values.
 
 examples:
   -
@@ -42,6 +41,11 @@ examples:
   -
     - Example Besselian Epoch time
     - |
+        !time/time B2000.0
+
+  -
+    - Example Besselian Epoch time, equivalent to above
+    - |
         !time/time
           value: 2000.0
           format: byear
@@ -53,23 +57,65 @@ examples:
           ["2000-12-31T13:05:27.737", "2000-12-31T13:06:38.444"]
 
   -
-    - Example array of times
+    - Example large array of times stored in a binary block
     - |
         !time/time
           value: !core/ndarray
-            data: [2000.1, 2000.2, 2000.3]
+            source: 0
+            shape: [1024]
             datatype: float64
+            byteorder: big
           format: jyear
+
+  -
+    - Example with a location
+    - |
+        !time/time
+          value: 2000.0
+          format: jyear
+          scale: tdb
+          location:
+            x: 6378100
+            y: 0
+            z: 0
+
+definitions:
+  iso_time:
+    type: string
+    pattern: "[0-9]{4}-(0[1-9])|(1[0-2])-(0[1-9])|([1-2][0-9])|(3[0-1])[T ]([0-1][0-9])|(2[0-4]):[0-5][0-9]:[0-5][0-9](.[0-9]+)?"
+
+  byear:
+    type: string
+    pattern: "B[0-9]+(.[0-9]+)?"
+
+  jyear:
+    type: string
+    pattern: "J[0-9]+(.[0-9]+)?"
+
+  yday:
+    type: string
+    pattern: "[0-9]{4}:(00[1-9])|(0[1-9][0-9])|([1-2][0-9][0-9])|(3[0-5][0-9])|(36[0-5]):([0-1][0-9])|([0-1][0-9])|(2[0-4]):[0-5][0-9]:[0-5][0-9](.[0-9]+)?"
+
+  string_formats:
+    anyOf:
+      - $ref: "#definitions/iso_time"
+      - $ref: "#definitions/byear"
+      - $ref: "#definitions/jyear"
+      - $ref: "#definitions/yday"
+
+  array_of_strings:
+    type: array
+    items:
+      anyOf:
+        - $ref: "#definitions/array_of_strings"
+        - $ref: "#definitions/string_formats"
 
 allOf:
   - tag: "tag:stsci.edu:asdf/0.1.0/time/time"
   - anyOf:
-    - type: string
-      # TODO: pattern
+    - $ref: "#definitions/string_formats"
 
-    - type: array
-      # We can't say `items: string` here, since then we'd couldn't
-      # have nested lists.
+    - $ref: "#definitions/array_of_strings"
 
     - $ref: ../core/ndarray
 
@@ -80,10 +126,10 @@ allOf:
             The value(s) of the time.
 
           anyOf:
-            - type: string
-            - type: number
-            - type: array
+            - $ref: "#definitions/string_formats"
+            - $ref: "#definitions/array_of_strings"
             - $ref: ../core/ndarray
+            - type: number
 
         format:
           description: |
@@ -192,9 +238,10 @@ allOf:
             Specifies the observer location for scales that are
             sensitive to observer location, currently only `tdb`.  May
             be specified either with geocentric coordinates (X, Y, Z)
-            with an optional unit or geodetic coordinates (longitude
-            in degrees, latitude in degrees, and optional
-            height).
+            with an optional unit or geodetic coordinates:
+              - `long`: longitude in degrees
+              - `lat`: in degrees
+              - `h`: optional height
 
           anyOf:
             - type: object
@@ -212,21 +259,21 @@ allOf:
               requiredProperties: [x, y, z]
             - type: object
               properties:
-                longitude:
+                long:
                   type: number
                   minimum: -180
                   maximum: 180
-                latitude:
+                lat:
                   type: number
                   minimum: -90
                   maximum: 90
-                height:
+                h:
                   type: number
                   default: 0
                 unit:
                   allOf:
                     - $ref: ../unit/unit
                     - default: m
-              requiredProperties: [longitude, latitude]
+              requiredProperties: [long, lat]
 
-      requiredProperties: value
+      requiredProperties: [value]
--- a/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
@@ -120,7 +120,9 @@ allOf:
 
     - $ref: "#definitions/array_of_strings"
 
-    - $ref: ../core/ndarray
+    - type: object
+      properties:
+        $ref: "../core/ndarray#anyOf/1/properties"
 
     - type: object
       properties:
--- a/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
@@ -62,14 +62,12 @@ examples:
           ["2000-12-31T13:05:27.737", "2000-12-31T13:06:38.444"]
 
   -
-    - Example large array of times stored in a binary block
+    - Example of an array of times
     - |
         !time/time
           value: !core/ndarray
-            source: 0
-            shape: [1024]
+            data: [2000, 2001]
             datatype: float64
-            byteorder: big
           format: jyear
 
   -
--- a/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
+++ b/schemas/stsci.edu/asdf/0.1.0/time/time.yaml
@@ -1,0 +1,232 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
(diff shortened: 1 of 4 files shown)


schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/0.1.0/time/time"
+title: Represents an instance in time.
+description: |
+  A "time" is a single instant in time.  It may explicitly specify the
+  way time is represented (the "format") and the "scale" which the
```
taldcroft: "which specifies ..."

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+                y:
+                  type: number
+                z:
+                  type: number
+              requiredProperties: [x, y, z]
+            - type: object
+              properties:
+                longitude:
+                  type: number
+                  minimum: -180
```
taldcroft: In astropy Longitude can be specified by any input value and it is then taken modulo 360 to have 0 <= value < 360.  Do these `minimum` and `maximum` represent hard limits on allowed values?

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+              UTC. For example, 946684800.0 in Unix time is midnight
+              on January 1, 2000.  [TODO: Astropy's definition of
+              UNIX time doesn't match POSIX's here.  What should we
+              do for the purposes of ASDF?]
+
+            - `cxcsec`: Chandra X-ray Center seconds from 1998-01-01
+              00:00:00 TT. For example, 63072064.184 is midnight on
+              January 1, 2000.  [TODO: Is this general enough to
+              include?]
+
```
taldcroft: What about support for the `FITS` format?  This allows specification of scale in a string.

http://astropy.readthedocs.org/en/latest/api/astropy.time.TimeFITS.html#astropy.time.TimeFITS

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+            - tt
+            - ut1
+
+        location:
+          description: |
+            Specifies the observer location for scales that are
+            sensitive to observer location, currently only `tdb`.  May
+            be specified either with geocentric coordinates (X, Y, Z)
+            or geodetic coordinates (longitude in degrees, latitude in
+            degrees, and optional height). [TODO: What are the units
```
taldcroft: In astropy the X, Y, Z are any distance quantities.  If you have a pick a unit for these in YAML I would go for `m`.

taldcroft:
I think the biggest conceptual issue here is how to support the full 128-bit precision in astropy.time.

On input this is relatively straightforward since you can add a `value2`:
```
        value2:
          description: |
            The secondary value(s) of the time which are added to the primary value(s).
            Default is 0.0.

          anyOf:   # Note: string not allowed
            - type: number
            - type: array
            - $ref: ../core/ndarray
```

The trickier bit is going the other way.  I wonder if it might make sense to special-case the `jd` format to always output `t.jd1` as `value` and `t.jd2` as `value2`.  In that way there would be a path for serializing a Time object to full precision.

There are other possibilities that would require some kind of option to serialize to high precision or not.   I.e. you could somehow choose to store `t.jd1` and `t.jd2` (as attributes `jd1` and `jd2` respectively) or `t.value` (default behavior now).

In any case users would also have the option of setting the precision in the Time object to 9 or 10 and that essentially gets full precision for a string-based output like ISO.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+
+            [TODO: Describing these scales in enough detail to
+            reimplement is perhaps too much to do here, but we should
+            have good ironclad references for each.]
+
+            The supported time scales are:
+
+            - `utc`: Coordinated Universal Time (UTC).  This is the
+              default time scale, except for the formats that
+              represents seconds since an epoch, namely `gps`, `unix`
```
taldcroft: The epoch_scale for `unix` is UTC, so that one (by chance) also defaults to UTC.  So maybe just say `This is the default time scale except for 'gps' and 'unix'`.

taldcroft:
Are you planning to do `TimeDelta` as well?

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+    and `location`.
+
+  - a string, in which case the format is guessed from across
+    the unambiguous options (`iso`, `byear`, `jyear`, `yday`), and the
+    scale is hardcoded to `UTC`.
+
+  In either case, a single time tag may be used to represent an
+  n-dimensional array of times, using either an `ndarray` tag or
+  inline as (possibly nested) YAML lists.  If YAML lists, the same
+  format must be used for all time values [TODO: This is a restriction
```
taldcroft: I think this would severely impact performance.  Basically now you pass the entire array to each Format parser and it will give an exception as soon as something fails.  If you don't assume that the inputs are homogeneous then every array element needs to be passed through the possible format parsers in turn to guess the format.  So for an input which is in the last attempted format the code would basically need to fail N x (M-1) times for N elements and M formats.  Instead now you have more like M fails until getting to the N successful parses.

(Yes, I suppose you could be clever and re-arrange the guess order on the fly..., but I think even this would add a lot of overhead).

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+
+            - `unix`: Unix time: seconds from 1970-01-01 00:00:00
+              UTC. For example, 946684800.0 in Unix time is midnight
+              on January 1, 2000.  [TODO: Astropy's definition of
+              UNIX time doesn't match POSIX's here.  What should we
+              do for the purposes of ASDF?]
+
+            - `cxcsec`: Chandra X-ray Center seconds from 1998-01-01
+              00:00:00 TT. For example, 63072064.184 is midnight on
+              January 1, 2000.  [TODO: Is this general enough to
```
taldcroft: It's probably the case that most people who deal with Chandra data would want to convert the time to a more generic format before serializing.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
```diff
@@ -0,0 +1,232 @@
...
+            - unix
+            - cxcsec
+
+        scale:
+          description: |
+            The time scale (or time standard) is a specification for
+            measuring time: either the rate at which time passes; or
+            points in time; or both. See also [3] and [4].
+
+            [TODO: Describing these scales in enough detail to
```
taldcroft: I think that http://www.iausofa.org/sofa_ts_c.pdf and references therein is sufficient.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: I see...  I'll think on that for a bit, but I think we can defer making a decision about that as far as ASDF is concerned for now...

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: That's what I was thinking, too.  I think I'll probably remove this one.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: I thought about including it, but I think YAML format provides a cleaner way to do this (by using a dictionary) and there's probably not much point to require ASDF libraries to implement a specialized parser for the FITS time scale convention.  As for the pyasdf implementation of this, it can save a FITS format time as a dictionary with the value (in iso) and scale.  I sort of have "reduce the amount of specialized parsing required" as a guiding principle in ASDF,  but if there's a good argument for including FITS scaled times verbatim, I could be convinced otherwise.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: Thanks.  That's helpful.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: Makes sense.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: Yes -- minimum and maximum are the limits (but you can also specify whether they are inclusive or not -- the default is inclusive).  I can certainly update the range to be 0 <= x < 360 -- or just leave it out.

mdboom:
> I think the biggest conceptual issue here is how to support the full 128-bit precision in astropy.time.

This is an important issue, and one I forgot to consider here.  I see the "pair of doubles" thing as primarily an implementation detail that arises out of a number of practical limitations.  Ideally, that implementation detail wouldn't leak up into the file format.  Maybe resolving #20 (to add extended precision formats to ASDF) would make sense to do first.  Or at least provide a way to standardize double-double values universally in the format.

> Are you planning to do TimeDelta as well?

Yes.  Maybe as a follow-on, though, as our immediate need is just for times.

schemas/stsci.edu/asdf/0.1.0/time/time.yaml
mdboom: One more thing about this -- the specific use case I've run into here is when converting some FITS files that have a mixture of `iso` and `isot`.  It seems that those parsers could be made less strict about the existence of the `T` without much loss in performance which might obviate a lot of this.

mdboom:
Giving this a little more thought -- I'm really resisting exposing the "double + double" representation to the level of the file format.

> In any case users would also have the option of setting the precision in the Time object to 9 or 10 and that essentially gets full precision for a string-based output like ISO.

As you say, it does seem to be the case that the ISO format, with enough decimals in the second part, is capable of representing the full precision, so it's not like full precision "can't be done" given the design here.

I'm in favor of:

- Adding a note to the standard that warns about precision limitations of the numerical formats and recommending `iso` or `yday` if more precision is required.

- The implementation can check if a single double is not sufficient to represent the time (basically by determining if the significand over/under flows) and if so either warn about loss of precision or just silently output `iso`.  But that's an implementation detail separate from the file format.

- Longer term, it should be possible to write a string-to-quad-precision parser to read the YAML, and convert the quad to two doubles before handing off to ERFA.  Requiring such a thing places a burden on ASDF library implementers, of course, that ISO date handling perhaps doesn't.

@taldcroft: Does that make sense -- am I missing something?

taldcroft:
@mdboom - that sounds fine.  Comments:

> The implementation can check if a single double is not sufficient to represent the time (basically by determining if the significand over/under flows) 

I don't exactly understand this because I think of precision as being a meta-attribute of a value (i.e. the uncertainty) which is not identifiable from the value itself.  For instance if I have a time which ends with 1/3 of a second then there is no way to know the precision a priori.  My feeling would be to just leave this up to the user and not try to be too clever.

> Longer term, it should be possible to write a string-to-quad-precision parser to read the YAML, and convert the quad to two doubles before handing off to ERFA

The existing parser for high precision ISO strings does produce the correct double-double JD representation and everything round-trips as expected.  This is built in to ERFA so any other implementation that uses ERFA gets this for free.  If an architecture has quad (float128) available and that is desired then just adding the two doubles from ERFA should work.  Without using ERFA then yes, a custom parser would be needed.

mdboom:
@taldcroft: You make a good point about precision being an attribute of the value -- i.e. something that the user needs to tell us.  My specific concern is that there is a precision loss "hole" that the user can fall through in the current design of astropy (or maybe I'm missing something).  I think this is getting away from what's relevant for this PR, since the plan is to say "if you need high precision, use ISO or yday", but this is the issue I was trying to avoid:

```
In [1]: from astropy import time

In [3]: t = time.Time(123456789123456789., 0.123456789123456789, format='jd', precision=9)

In [4]: str(t)
Out[4]: '1.23456789123e+17'
```

In the above, since I specified the maximum precision, I would have expected the string to preserve all the precision of the input (or at least up to the precision available in the pair of doubles).  For ASDF, I was suggesting that this would fall back to an ISO representation here, since there is no standard for a "pair of doubles" string.  In astropy, this might be surprising, but at least there is a solution for users to get around it.  For ASDF, where at a minimum we have to ensure things round trip losslessly, we would have to deal with this automatically.  Anyway, all is moot, since I think I'm going to not include the "two doubles" representation in astropy anyway (or at least punt it into the future).

taldcroft:
OK, I understand your issue now.  The `precision` attribute is confusing here because it only applies to the precision for an ISO-like representation.  In other words there are ERFA routines that convert a double-double time to ISO and honor the `precision`.  In the case of outputting the string representation of a Time object in JD, that just adds the two doubles and stringifies it.  In theory we could make a custom method to output a string corresponding to the two doubles, but I'm not sure we want to go there.

Note that just going to ISO is not a complete solution, since your example gives an error in ERFA:
```
>>> t.iso
...
ErfaError: ERFA function "d2dtf" yielded 1 of "unacceptable date (Note 6)"
```
The takeaway is that currently the astropy Time object does not support a text representation of the object that always round trips.  That's kinda why I was suggesting going down the path of 2 double arrays, but maybe we should think more about the representation issue?

taldcroft:
Just thinking some more, it seems the only way to *guarantee* lossless representation is either two-doubles or a quad (either binary or written out in full precision).  Since the underlying data are two doubles, then any other serialization via an existing astropy time format is subject to loss.

Issue #90, opened 2015-06-26
Documentation on what ASDF format versions mean


embray:
In the introduction to the file layout, the documentation for the ASDF [header](http://asdf-standard.readthedocs.org/en/latest/file_layout.html#header) explicitly breaks the ASDF version into a "major", "minor", and "micro" version.  But we don't really discuss how versions will be used for the format.  Since there is currently only one version it may yet be premature to say much about it, so I'm putting this here as a reminder.

In particular I think we need to think about how the format version relates to the ASDF standard and to ASDF software.  Of course, there's also room for flexibility and change in how we interpret versions (to that end  we may even want to think ahead an add an optional "epoch" field to the version specification).

Also, I don't think the version of the _standard_, as a document, and the version of the format are always necessarily identical either.  For example, we want to be able to release updates to the standard that patch typos or include clarifications or fill in missing information, but that do not actually change _anything_ about the format or even the schemas.  Though in the interest of keeping things sane I think we still want to keep the standard version in sync with the format version (for example one type of bug fix might be to patch a missing or incorrect rule in one of the schemas, which amounts to an update both in the format and the standard).  To that end I think the version of the standard may also include `.post<N>` tags for minor text updates that don't change the format.

How the format version relates to software versions is trickier and may benefit from experience.  I definitely don't think the two should be tied in any way--the version of an ASDF library should not in any way be tied to what ASDF versions it supports (although that should be well documented).  More importantly will be determining when updates to the ASDF _format_ require updated software.  My feeling is that ASDF files written with newer software should at least be readable by older software the supports the same _major_ version.  In other words, an increase in _minor_ version may have some benefits, but should still be backwards compatible with older software.  Only increases in _major_ versions may introduce backwards incompatibility with older software.  I think that's a good, conservative guideline.  Though I'm not 100% sure how it will be in practice...

mdboom:
I agree with this -- particularly that the library and the specification versions should not be tied to one another.  I think it's not too premature to develop at least some aspirational language about this.

Is this a fair summary for the version of the standard itself...?

- major: Introduces a backward incompatible change.
- minor: Introduces a substantial change which does not break backward compatibility.  i.e. libraries that don't understand the new changes can still read the file, but may not be able to act on the new features, though all old functionality would still work.
- micro: A minor change or clarification to the text of the standard that doesn't impact the file format itself.  (This means that the micro number in as ASDF file should basically always be ignored by a library).

[...]

PR #91, opened 2015-07-01
Add versioning conventions

mdboom:
This is an attempt to address #90.

This is very much up for comment -- I just thought it would be helpful to put up some language to start hacking away at.

Cc: @embray @perrygreenfield 

TODO:

- [x] Once we've agreed on the conventions, the standard will need to be updated to follow them
- [x] I think this necessitates the need for an "index" file that would map versions of the ASDF standard to the most recent version of the file format and each of the schemas.  Also to have a map that maps each `major` and `major.minor` version to the most recent concrete example in that set.  i.e. `1 -> 1.2.1`, `1.1 -> 1.1.0`, `1.2 -> 1.2.1`.


source/versioning.rst
```diff
@@ -0,0 +1,92 @@
...
+  conventions of the most recent understood version.
+
+- When encountering a **minor version** that is greater than the
+  understood version, a warning should be emitted, and the library
+  should attempt to handle the element using the conventions of the
+  most recent understood version.
+
+- When encountering a **patch version** that is greater than the
+  understood version, silently ignore the difference and handle the
+  element using the conventions of the most recent understood version.
```
mdboom: This makes me wonder whether patch numbers should be in the file itself at all if they are always ignored.

source/versioning.rst
```diff
@@ -0,0 +1,94 @@
...
+
+When writing ASDF files, it is recommended that libraries provide both
+of the following modes of operation:
+
+- Upgrade the file to the latest versions of the file format and
+  schemas understood by the library.
+
+- Preserve the versions of the file format and schemas of the input
+  file.
```
mdboom: This latter one is going to require some rejiggering of how pyasdf works, and has some of the same problems of "preserving comments" which is "where to you attach the metadata to help you remember what version to write something back out as".  But if we all agree it's desirable, we should try to find a solution.

source/versioning.rst
embray: I was wondering that myself.  I lean in favor of including it for now (even if in many cases it's just left as zero).  As I wrote when I first raised this issue, this is a policy that will likely have to be adjusted as experience dictates.  So I would definitely want to leave open the possibility of having patches to the file format itself.

source/versioning.rst
embray: It definitely opens a can of worms.  I think preserving the version of the file format itself is probably not a problem, but the versions of the schemas sounds tricky.  Maybe this option should only be available (in fact I would argue required) for in-place updates to an existing file (as opposed to reading a file into native data structures, then writing out to an entirely new file).  One option would be to allow specifying all the relevant versions when writing a new ASDF file.  This would allow dumping a list of versions from an existing file and storing that off in a variable for later use.  There's also the issue of error handling when some object *can't* be stored given the list of format / schema versions to use.

embray: In other words, maybe it's best to punt on that for now.  When I first suggested versioning all the things for the new format in the first place it was to enable growth when needed (out of frustration with FITS' lack of any sensible way to do this).  It never meant that all issues of file format translation had to be resolved up front.

source/versioning.rst
mdboom: Just so I make sure I understand you, by definition, a patch number should never be anything that a library acts on -- if it were, it should be a minor or major number change.  So, strictly speaking, it's unnecessary to write the patch numbers into as ASDF file (either for the file format or for a schema version).  The question is whether *as human documentation* they should be included anyway -- or as insurance against needing them later if this versioning policy changes in the future.  They would, of course, always be included in the specification.

source/versioning.rst
mdboom: > Maybe this option should only be available (in fact I would argue required) for in-place updates to an existing file (as opposed to reading a file into native data structures, then writing out to an entirely new file).

The present problem -- which is resolvable -- is that `pyasdf` always reads the file into native data structures.

> One option would be to allow specifying all the relevant versions when writing a new ASDF file. This would allow dumping a list of versions from an existing file and storing that off in a variable for later use. 

This seems more tenable.  It also would allow for saying "save to ASDF X.Y.Z", which would then look up in the version index which versions of the schemas were the latest shipped with that specific ASDF version and use that.  That seems like a pretty common use case.

This is perhaps an argument for storing the ASDF spec version explicitly in the file.  (In the first bullet point about the "types of versions" above, I say it isn't).  Let's say the library was operating in "version preserving" mode: If you don't include the ASDF spec version in the file, a library would have to "guess" based on the versions of schemas in the file which version to write back out as.  If you include it, it's trivial -- you can use the ASDF spec version to look up in the version index what versions of everything else to use.  So maybe we modify the header of the ASDF file to be:

```
#ASDF A.B.C X.Y.Z
```

where `A.B.C` is the specification version and `X.Y.Z` is the file format version.

source/versioning.rst
embray: I think we're in agreement.  I mostly would want to have the patch number in the format (even if it's kept as zero), as you say, as "insurance against needing them later if this versioning policy changes in the future"

source/versioning.rst
embray: Wouldn't this then just be an argument for keeping the format version and the spec version the same?  I think having two versions could get overly confusing.

mdboom: The ASDF spec could have the major version bumped when the major version on the ndarray schema is bumped.  That wouldn't require a major version bump to the file format layout, and the file would still be readable (with the exception of any new-style ndarrays) by an older library.

embray: Maybe, then, core schema versions *should* be decoupled from the format version except where necessary (example, if some major change to the ndarray schema also requires a change to the file layout or processing in some way).

mdboom:
> Maybe, then, core schema versions should be decoupled from the spec version except where necessary (example, if some major change to the ndarray schema also requires a change to the file layout or processing in some way).

The spec version should be incremented any time anything is changed (and I don't think you're arguing against that)...  You just saying that if the major version of ndarray is bumped, the major version of the spec doesn't have to be bumped.  That perhaps makes some sense.

> In other words, I'm unless the format version and the spec version are the same, I'm not sure what would drive forward the format version.

I still think (even with the change immediately above) the format and the spec format need to be decoupled.  A new version of the spec may be released with a whole new set of transformation types (for example) that wouldn't require any change in the underlying format.

In any event, I've updated this to reflect some of this feedback.

@perrygreenfield: Any thoughts?  I feel like there's a line between complexity and flexibility here that we need to carefully walk.

source/versioning.rst
perrygreenfield: Wouldn't it be more semantically clear if it didn't include the patch number when it didn't affect the code, and then if that policy changed in the future, it then was added? Not there, it doesn't matter, if there, it does.

perrygreenfield:
Where is the previous discussion on this? I think I'm missing some context.

source/versioning.rst
mdboom: Yeah -- I can see this both ways.  Code that expects `X.Y` may just use `float('X.Y')` and then break if we change it to `X.Y.Z`.  But maybe I'm just being overly cautious there.  Don't feel strongly either way.

mdboom:
#90 has some previous discussion.  Don't know if it goes back any further than that.

perrygreenfield:
Hmmm, if new transforms are added, that sure seems to suggest it may impact the required software to read them, wouldn't it? Yes, older files are not affected, but it does mean that there are new cases that require new software to interpret, doesn't it?

If WCS was some sort of plugin, then, no, it wouldn't require a new core version but to use WCS it may require a new plugin version. But I'm not sure I'm interpreting the versioning dichotomy correctly (we are almost talking about 3: format, specification, software). 

Based on the previous discussion, it seems that specification patches could, by definition, be ignored as far as the format goes. But any other kind could stay in lockstep with the format (are there any examples where that isn't true?)  The transform case doesn't seem to be one of them.

mdboom:
> Hmmm, if new transforms are added, that sure seems to suggest it may impact the required software to read them, wouldn't it? Yes, older files are not affected, but it does mean that there are new cases that require new software to interpret, doesn't it?

Yes -- however, if those new transforms aren't used in a particular file, older software could still read it.  That's the reason to version the schemas separately from the spec as a whole.

> But I'm not sure I'm interpreting the versioning dichotomy correctly (we are almost talking about 3: format, specification, software).

In the current iteration of this PR, we're actually talking about 4: the specification as a whole, the low-level file format, the schemas within it, and the software.  All would have independent versions (though the software would probably advertise compatibility with a particular version of the specification).

> Based on the previous discussion, it seems that specification patches could, by definition, be ignored as far as the format goes. But any other kind could stay in lockstep with the format (are there any examples where that isn't true?) The transform case doesn't seem to be one of them.

The transform example is an example of a change that only matters if you use the transform features -- it has no impact on you if you're only using ndarray, for example.  Remember, we already have "modules" of schemas with explicit dependencies.  



It might make sense to version these modules (but not the individual schemas within).

source/versioning.rst
embray: I'm mostly fine either way.  We could make the final `.0` optional, just so long as the software expects that there *may* be additional version info after that.  This is the one bit that has to be somewhat future proof, because even older ASDF readers have to be able to read just this one line of text well enough to be able to say "This version of ASDF is not supported yadda yadda."

embray:
I certainly agree with where @mdboom is going with respect to the file version--the file version simply indicates to the reader whether the informational content of the file (all data and metadata) can be correctly *parsed* by the reader.  However, the semantic interpretation of that information, once read, may depend on what schemas the reader supports, and what version(s) of those schemas.

This is somewhat akin to existing FITS conventions--for example one can read the binary table containing a compressed image perfectly fine, but without support for the tile compression convention there's no way to further interpret that data.

The difference in ASDF is that it is explicit what data structures are used within a given file, including a version if there have been updates to their format, and a schema to validate use of that format.  And because this is part of the standard, even software that has no prior knowledge about a given schema can still report that it's in use by that file, and provide information about it.  Another difference is perhaps merely semantic--the WCS component is not a mere "convention"--it is very much a core component of the standard.  But this design still allows implementation of it by readers to be optional, in cases where full WCS support is not necessary.

embray:
> It might make sense to version these modules (but not the individual schemas within).

Will have to think on that... I keep going back and forth between whether I like it or hate it :)

Part of me also wants both, but then any time you update one of the schema's versions the "module" version will also have to be updated making the individual schema versions unnecessary.  At the same time, users should be able to introduce individual schemas for their data models without it having to be a part of some "module" containing multiple schemas.

mdboom:
This now (I think) includes feedback from @perrygreenfield, @embray and @mdboom's discussion yesterday.  Let me know if I missed anything or can further simplify.

embray:
I like the third-person self-reference :)

source/file_layout.rst
```diff
@@ -42,25 +44,33 @@ All ASDF files must start with a short one-line header.  For example::
...
+    version of the ASDF specification, and is only updated when a
+    change is made that affects the layout of file.  See
+    :ref:`versioning-conventions` for more information about these
+    versions.
 
-  asdf_token = "#ASDF"
-  major      = integer
-  minor      = integer
-  micro      = integer
-  header     = asdf_token " " major "." minor "." micro ["\r"] "\n"
```
embray: I like the clarification of what the file format version means, but I don't understand what it's been separated as not part of the asdf_token.  Or perhaps we could call the token you have described above as the `asdf_header` which is made up of the `asdf_token` and `format_version`.  I think it was nice to have the parsing described here (even though we later mention that it follows semver).

mdboom: Yeah -- I should probably revert most of this back to the original...

source/versioning.rst
```diff
@@ -0,0 +1,104 @@
...
+- **File format version**: Refers to the version of the blocking
+  scheme and other details of the low-level file layout.  This is the
+  number that appears on the ``#ASDF`` header line at the start of
+  every ASDF file and is essential to correctly interpreting the
+  various parts of an ASDF file.
+
+- **Schema versions**: Each schema for a particular YAML tag is
+  individually versioned.  We encourage schemas provided by third
+  parties (i.e. not in the ASDF specification itself) to be versioned
+  as well.
```
embray: "This allows schemas to evolve while still allowing data written to an older version of that schema to be validated correctly."

embray: (Just a suggestion--I think it clarifies the question of "Why should I use a version on my schema?"  I think it's obvious, but I would still anticipate the question.)

source/versioning.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Handling version mismatches
+---------------------------
+
+Given these conventions, the ASDF standard recommends certain behavior
+of ASDF libraries.  ASDF libraries should, but are not required, to
+support as many existing versions of the file format and schemas as
+possible, and use the version numbers in the file to act accordingly.
+
+For future proofing, the library should gracefully handle version
```
embray: Super minor nitpick but "future-proofing" should be hyphenated.

source/file_layout.rst
mdboom: On looking at this again, I think it best we just say "the version number conforms to the semantic versioning specification 2.0.0".  It's actually more complex than what we had before (particularly the different kind of suffixes that may appear), so I don't know if I want to repeat all that detail here.   Though let me know if you have strong objections -- there may be a middle road.

source/versioning.rst
mdboom: Good point.  Added.

mdboom:
Aside from my comments today, are there any further suggestions?  Moving to this will require some significant grep/sed/moving magic that I'd prefer to only do once ;)

source/file_layout.rst
embray: Maybe just keep
```
asdf_token = "#ASDF"
header       = asdf_token " " format_version ["\r"] "\n"
```
where the syntax for `format_version` is described in the semver spec?

Issue #94, opened 2015-07-14
Required property for ASDF writer library


embray:
Directly related to #3, I think somewhere in the top level of the tree we should *require* a property indicating the software (name and version, at least) that wrote out that file.  Requiring this I think would at least discourage ad-hoc writing of files.  I think a timestamp should go along with this.  This wouldn't help much if users then edit a file by hand, but at least then the timestamp will be different from the file's last modified time (unless the user manually edits the timestamp too, but even then it will probably be wrong).

Maybe if we *really* wanted to be strict we could require that every ASDF writer have a unique signature it uses to sign the file.  This doesn't need to be cryptographically secure or anything--just something good enough to confirm that the file was indeed written by that software (or some knucklehead manually computing the signature using some library's key and updating it by hand).

The point is to have some idea who to blame if invalid ASDF files are being written :)

mdboom:
I like the idea of the required property with the software and version (and maybe a URL to the homepage of the software).  That would certainly have been helpful in VOTable land (where an XML comment with the software name is *encouraged* but not *enforced*).

I'm a little more concerned about a timestamp.  It makes in-place updating of the data blocks more difficult (or at least means that the timestamp would only be good enough to indicate file creation not modification -- or maybe it only indicates modification of the tree).  Or the timestamp could be outside of the tree and of fixed length (maybe right after the `#ASDF` header), in which case it could be edited in-place.

Even a checksum of the YAML content would be good enough to detect if the file was edited by hand (again, assuming someone doesn't go through the trouble of computing it by hand also).  That would allow a library to throw up a big fat warning that "all bets are off", but then still try its best anyway.

And, if I'm following that this idea also somewhat follows from the problems outlined in astropy/astropy#3941, one solution is also to establish a "convention" early on that code to workaround invalid files will not be written.  (Of course, in the real world, I would guess that the spec is probably underspecified in a lot of ways and some things will be invalid "by accident", and we'll have to make clarifications as we go on.  In which case, we're still better off then FITS because we have a version number and can deprecate things).

embray:
My idea about adding a signature using a per-library random key was that the signature would even be checked (in most cases at least) when reading the file.  It's more for debugging purposes--to determine "Was this file really written by that software?"  It's true though that could be a problem for things like in-place updates, and especially if the file is being updated by different software than the one that wrote it.  Need to think about that more...

embray:
[...]

PR #101, opened 2015-07-22
Add schema for describing software

mdboom:
This partially addresses #94, and provides a location to describe the ASDF-writing library that produced the file, as well as any other software that may have been involved.

This doesn't address the verification/checksumming features discussed in #94.  I think that's best left as a separate PR.


embray:
Ah, I like this.  I had thought of maybe having a history of all the different software that touched a file, but I wasn't sure how to manage that in a reasonable way.  Should've thought bigger.

schemas/stsci.edu/asdf/0.1.0/core/asdf.yaml
```diff
@@ -11,6 +11,20 @@ description: |
...
+    $ref: software
+
+  software:
+    description: |
+      Lists any relevant software that produced or modified the ASDF
+      file.  May include such things as data calibration pipelines,
+      data analysis tools etc.
+    type: array
+    items:
+      $ref: "software"
```
embray: What if we instead turned this into a full "history" log, including the software that made the change, the timestamp, an an optional log message (which would be passed in by the user via the API when saving or updating a file).
Could include other optional log metadata but I'm not sure how crazy to go with this.

schemas/stsci.edu/asdf/0.1.0/core/software.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+  version:
+    description: |
+      The version of the software used.  It is recommended, but not
+      required, that this follows the (Semantic Versioning
+      Specification)[http://semver.org/spec/v2.0.0.html].
+    type: string
+
+requiredProperties: [name, author, homepage, version]
+additionalProperties: true
```
embray: :+1:

schemas/stsci.edu/asdf/0.1.0/core/asdf.yaml
embray: Or to put another way, as defined this is a little ambiguous: Should it be a set, excluding uniques, of all the software that touched the file?  Does the order of the entries mean anything?  

I think by making this into the official history log of the file the intent is clearer, in that it will list each software any number of times each time it updated the file.  Having this baked into the file format could be a big win for provenance too.

mdboom: Yeah -- a history log makes sense.  (For the `software` list, but not the `asdf_library` entry, for which there can only ever be one that wrote the file last -- but maybe that's obvious).

mdboom: I think if this becomes more like a log, in addition to the software, it each entry should have a free form description along the lines of "applied cosmic ray rejection".

mdboom:
Updated with some of @embray's ideas.

schemas/stsci.edu/asdf/0.1.0/core/asdf.yaml
embray: Oh yes, definitely.  That's what I meant that it should have an optional log message passed in by the user (where "user" in this case can be other software of course).

embray:
Very cool.  I guess one loophole to this is that if a file is hand-edited it won't necessarily get the history added properly.  But that's something for users to be aware of then.  Maybe not too bad since I've heard of people hand-editing HISTORY keywords in FITS files too...

schemas/stsci.edu/asdf/0.1.0/core/history_entry.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+  time:
+    description: |
+      A timestamp for the operation.
+    type: string
+    format: date-time
+
+  software:
+    description: |
+      A description of the software that performed the operation.
```
embray: Although now I have a question about this--is this the ASDF library that produced the file, or is this meant to be some other software that assembled or modified the file (regardless of what ASDF library it's using)?  I assume the latter, which makes sense.  But I wasn't thinking clearly about that at first.  So in that case `asdf_library` would just be the *last* ASDF library that updated the file?

mdboom:
We could add a history checksum?

embray:
Well, I guess that gets back to the checksumming issue (which I agree can be separate).  I don't see why the history would need a checksum if the overall file has a checksum?

mdboom:
That's true.  We don't want/need a proliferation of checksums.

schemas/stsci.edu/asdf/0.1.0/core/history_entry.yaml
mdboom: The `asdf_library` member is always the last library that updated the file.  Since -- other than updating binary data in place, I suppose -- it's sort of irrelevant what the previous ASDF library was.

The `history` list is meant to be a history of modifications to the file and to include "higher-level" software, such as calibration pipelines.  That does raise the question of how much software detail to include -- for example, a JWST calibration pipeline will be using a particular version of astropy, numpy, Python, OS etc..  We could include those here (and thus make `software` a list), but I worry that could get pretty out-of-hand.

embray: Presumably the amount of detail to include in this list would be up to the user.  For example a program that updates the ASDF file is free to write an entry into the history about itself.  But there's nothing stopping it from including however much detail it wants in the description field.  I agree there may be such a thing as "too much" here, but we don't have to decide that right now.

mdboom: True -- I guess the only thing to decide right now is whether each entry in the history list should have a single software entry or a list of software.  Then it would still be up to each program how much detail about its dependencies it wants to include, but it would be slightly more structured than just cramming all of that in the description field.

embray: I wouldn't be opposed to changing it something like
```
oneOf:
    - $ref: software
    - type: array
      items: {$ref: software}
```
That's pretty harmless--if someone wanted to use multiple they could, but they don't have to either.

mdboom: Yeah, that works.

PR #104, opened 2015-07-22
Describe problem with large numeric literals

### LINKED ISSUE #88: Should be more specific about precision of numeric literals in tree
Reading of numeric literals in YAML happens differently in different YAML parser implementations.

For example: PyYAML will parse an arbitrary length integer into a Python long, so they are truly arbitrary-sized integers.  js-yaml, however, parses into a float64, since everything is a float64 in JavaScript.  Reading a file produced by Python into Javascript would result in unexpected data loss.  To be truly portable, an integer should use no more than 52 significant bits (since that's the size of the significand in a double).

There's potential differences with floating-point numbers as well, though everything I've tried seems to be float64/double.  (The YAML spec seems to be silent on this, AFAICT).  But as hinted in #20 and for the time specification, it would be nice if this supported quad precision someday if possible.

So what should the spec say about this?  That the significant bits should be no more than 52?  (Meaning trying to write out a large integer from Python would raise an exception).  Or should it just be a recommendation (meaning perhaps a warning about this when writing out a large integer).  Or, we could explicitly say integers are arbitrary length, in which case languages that don't currently parse YAML that way would be forced to grow such a feature to be truly ASDF compliant.

- [ ] Add a reference file test with large integer and floating-point literals

Cc: @embray, any thoughts?

Related to #20.

mdboom:
Fix #88


embray:
:+1:

embray:
(I wouldn't bother saying this in the standard, but if someone had an immediate need for higher-precision numbers they can always make a custom tag + schema for it.)

mdboom:
@embray: That's true.

PR #106, opened 2015-07-30
Update wcs schemas to reflect changes in gwcs

mdboom:
Cc: @nden


mdboom:
Some of the descriptions here aren't terribly -- uh -- descriptive.  @nden: May need your help with some of that.

schemas/stsci.edu/asdf/wcs/wcs-0.1.0.yaml
```diff
@@ -9,20 +9,24 @@ description: >
...
       A list of steps in the forward transformation from detector to
       world coordinates.
-
       The inverse transformation is determined automatically by
       reversing this list, and inverting each of the individual
       transforms according to the rules described in
-      [inverse](ref:http://stsci.edu/schemas/asdf/transform/transform-0.1.0/anyOf/0/properties/inverse).
-
-    $ref: "steps-0.1.0"
+      [inverse](ref:http://stsci.edu/schemas/asdf/0.1.0/transform/transform/anyOf/0/properties/inverse).
```
nden: Is this true (given that this is a list of steps)? It's correct for each step but the entire `inverse` must be constructed by chaining the inverse of each step in the reverse order, right?

mdboom: Yes -- you're right.  But how is that different from what is here (or how can it be worded less confusingly)?

nden: Oh, yes. I don't know what I was thinking (I guess I wasn't).

schemas/stsci.edu/asdf/wcs/celestial_frame-0.1.0.yaml
```diff
@@ -0,0 +1,30 @@
...
+title: >
+  Represents a celestial frame.
+
+allOf:
+  - type: object
+    properties:
+      reference_position:
+        description: |
+          The position of the reference frame.
+        enum: [geocenter, barycenter, heliocenter]
```
nden: Remove `reference_position` from `celestial_frame` and add it to `spectral_frame`.

schemas/stsci.edu/asdf/wcs/frame-0.1.0.yaml
```diff
@@ -0,0 +1,47 @@
...
+    items:
+      anyOf:
+        - type: string
+        - type: 'null'
+
+  reference_frame:
+    description: |
+      The reference frame.
+    enum: [ICRS, FK5, FK4, FK4_noeterms, galactic, galactocentric,
+           altaz, GCRS, CIRS, ITRS, precessed_geocentric]
```
nden: One issue with the frame schema is that `reference_frame` is written as a string and does not support objects. Reference frames (other than ICRS) need attributes to be fully defined. Is there a chance that this can support the frame_attributes dict in astropy.coordinates? I think reference_frame only needs to have additional properties allowed where the frame attributes can be dumped. I see Time already has its own definition.

schemas/stsci.edu/asdf/wcs/step-0.1.0.yaml
```diff
@@ -6,32 +6,24 @@ tag: "tag:stsci.edu:asdf/wcs/step-0.1.0"
...
       final output axes.
-    $ref: "../transform/transform-0.1.0"
+    anyOf:
+      - $ref: ../transform/transform-0.1.0
+      - type: 'null'
+    default: null
 
-required: [name]
\ No newline at end of file
+required: [frame]
```
nden: `transform` is required as well, right? Even if it is null.

mdboom: I think it's fine to have it not required (as it definitely isn't required in the last frame).  It has a default of `null` in any case.

mdboom:
@nden: Do you think this is ready to merge?  (This is the last thing for asdf 1.0)

nden:
:+1: to merge this one too.

PR #107, opened 2015-07-30
Fix default values in projections

mdboom:


mdboom:
See astropy/astropy#4021

PR #108, opened 2015-08-20
Show raw YAML on rendered schema pages, in addition to raw download link

embray:
Instead of providing just a download link on schema pages, show the original YAML with syntax highlighting *and* provide a raw download links. 

For example:
asdf-schema-page

### DIFF
--- a/convert_schemas.py
+++ b/convert_schemas.py
@@ -370,8 +370,12 @@ def convert_schema_to_rst(src, dst):
         name += ': ' + schema['title'].strip()
     recurse(o, name, schema, [id], 0)
 
-    o.write(".. only:: html\n\n   :download:`Original schema in YAML <{0}>`\n".format(
+    write_header(o, 'Original schema in YAML', 1)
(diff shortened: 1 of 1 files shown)


embray:
I thing this is still useful to display in the docs, just to get people accustomed to looking at YAML if nothing else.

mdboom:
This is probably a good idea.  (And there are some corner cases where the JSON schema usage is a bit odd and doesn't really "convert" well to my ad-hoc human readable form, so having the YAML original is helpful).

And a question (I don't really have a strong opinion, I think): Should this also be included in the PDF, or should this be HTML only?

embray:
> (And there are some corner cases where the JSON schema usage is a bit odd and doesn't really "convert" well to my ad-hoc human readable form, so having the YAML original is helpful).
That too. I still like it in general though, so it will be nice to tweak the areas where it doesn't work as well, at some point.

> Should this also be included in the PDF, or should this be HTML only?
I could also go either way.

mdboom:
Ok -- let's merge as-is.  We can always remove from PDF later.

PR #117, opened 2015-12-01
Add lookup_table schema

nden:
Should this schema be versioned 1.0.0 or 1.1 or something else? @embray Do you know?


embray:
Hmm, good question.  I'll have to look over my notes on that.  I think last we discussed this we decided for now that schemas making up the core schemas should be in lock step with the format version.  But we're not really updating the format by adding this schema, I don't think.  So for now I think it can be "1.0.0" and we'll revisit that question later.

nden:
Superseded by #122

PR #122, opened 2016-06-17
Add Tabular model  schema

nden:
This is the schema for `astropy.modeling.tabular.Tabular`. Should be merged after astropy/astropy#5105


nden:
@justincely Since you are the last ine who did changes and actually got the tests to pass. Any idea why Travis is failing? Looks like a pdf is not building.

justincely:
@nden I've merged into master a change that gets the travis build to pass.  It was just failing on the PDF document creation, which isn't necessary anyway.  I'll look into a broader cleanup of the docs, but this will suffice for you now.

PR #142, opened 2017-06-08
Update wcs schemas to reflect astropy changes

drdavella:
Astropy changed its representation of the ``obsgeoloc`` and ``obsgeovel``properties from ``Quantity`` objects to ``CartesianRepresentation`` objects.

This commit adds a new ``Quantity`` schema since each dimension of a ``CartesianRepresentation`` can be represented as a ``Quantity`` with a unique value and unit. It updates the ``wcs`` frame schema to represent a ``CartesianRepresentation`` as an array of ``Quantity`` objects.

### DIFF
--- a/schemas/stsci.edu/asdf/wcs/frame-1.0.0.yaml
+++ b/schemas/stsci.edu/asdf/wcs/frame-1.0.0.yaml
@@ -168,15 +168,13 @@ properties:
           when `reference_frame` is `GCRS` or `precessed_geocentric`.
         type: array
         items:
-          - type: array
-            items:
-              type: number
-            minItems: 3
-            maxItems: 3
-          - $ref: ../unit/unit-1.0.0
-            default: m
+          $ref: quantity-1.0.0
+        minItems: 3
+        maxItems: 3
         default:
-          - [0, 0, 0]
+          - { value: [0], unit: m }
+          - { value: [0], unit: m }
+          - { value: [0], unit: m }
 
       obsgeovel:
         description: |
@@ -186,15 +184,13 @@ properties:
           when `reference_frame` is `GCRS` or `precessed_geocentric`.
         type: array
         items:
-          - type: array
-            items:
-              type: number
-            minItems: 3
-            maxItems: 3
-          - $ref: ../unit/unit-1.0.0
-            default: m/s
+          $ref: quantity-1.0.0
+        minItems: 3
+        maxItems: 3
         default:
-          - [0, 0, 0]
+          - { value: [0], unit: m/s }
+          - { value: [0], unit: m/s }
+          - { value: [0], unit: m/s }
 
     required: [type]
 
--- a/schemas/stsci.edu/asdf/wcs/quantity-1.0.0.yaml
+++ b/schemas/stsci.edu/asdf/wcs/quantity-1.0.0.yaml
@@ -1,0 +1,27 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/wcs/quantity-1.0.0"
+
+title: >
+  Represents a Quantity object from astropy
+
+anyOf:
+  - tag: "tag:stsci.edu:asdf/wcs/quantity-1.0.0"
+  - {}
+
+type: object
+properties:
+  value:
+    description: |
+      A vector of one or more values
+    type: array
+    items:
+      type: number
+    minItems: 1
+  unit:
+    description: |
+      The unit corresponding to the values
+    $ref: ../unit/unit-1.0.0
+required: [value, unit]
+...


nden:
A few initial comments on this PR - it's good to have a schema for `Quantity`. I would move it under `unit` though (at least I'd look for it there because it's under unit in astropy). 

Also I think the versions need to be `quantity-1.1.0` and `frame-1.1.0`  not 1.0.0. I don't have first hand information on the intended versioning implementation but that's how I am interpreting the description 
[here](http://asdf-standard.readthedocs.io/en/latest/versioning.html) and in #90 and #91.
The schemas should be added to the corresponding version_map files under schemas/stsci.edu/asdf.

Naively I would have created also a corresponding tag in asdf/asdf/tags/unit, similar to the `Unit` class.
I think some of the code in test_wcs.py will go there - it needs to implement `to_tree` and `from_tree`.
Add a few test cases which roundtrip simple quantities.
(As a side note, `Quantity` is a subclass of `np.ndarray`, so I am not sure how asdf will recognize them. Currently it saves a Quantity instance as a numpy array and doesn't roundtrip.)

schemas/stsci.edu/asdf/wcs/frame-1.1.0.yaml
```diff
@@ -115,24 +115,16 @@ properties:
...
+          The ICRS coordinates of the Galactic center. Required when
+          `reference_frame` is `galactocentric`.
+        $ref: icrs_coord-1.0.0
 
       z_sun:
         description: |
           The distance from the sun to the galactic midplane.  Required
-          when `reference_frame` is `galactocentric`.  Required when
+          when `reference_frame` is `galactocentric`.
           `reference_frame` is `galactocentric`.
```
drdavella: Left over from copy/paste, should be removfed.

schemas/stsci.edu/asdf/wcs/icrs_coord-1.0.0.yaml
```diff
@@ -0,0 +1,40 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/wcs/icrs_coord-1.0.0"
```
nden: Should this be 1.1.0 since 1.0.0 is released already?

schemas/stsci.edu/asdf/wcs/spectral_frame-1.1.0.yaml
```diff
@@ -0,0 +1,30 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/wcs/spectral_frame-1.1.0"
```
nden: Not sure 1.1.0 is needed.

schemas/stsci.edu/asdf/wcs/step-1.2.0.yaml
```diff
@@ -0,0 +1,31 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/wcs/step-1.2.0"
```
nden: Since 1.1.0 has not been released isn't it ok to accumulate changes and not bump to 1.2.0?

schemas/stsci.edu/asdf/wcs/celestial_frame-1.1.0.yaml
```diff
@@ -0,0 +1,25 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/wcs/celestial_frame-1.1.0"
```
drdavella: I might be misunderstanding it as well. The reason I bumped the version here and in other places is because these schemas now make reference to `frame-1.1.0` whereas previously they used `frame-1.0.0`. Maybe they should just reference a generic `frame` though (without a specific version) and then a version bump won't be necessary?

schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
```diff
@@ -0,0 +1,27 @@
...
+title: >
+  Represents a Quantity object from astropy
+
+anyOf:
+  - tag: "tag:stsci.edu:asdf/unit/quantity-1.1.0"
+  - {}
+
+type: object
+properties:
+  value:
```
drdavella: Yes, that makes sense. I'm not sure whether `ndarray` is automatically handled or not but I'll investigate.

schemas/stsci.edu/asdf/wcs/icrs_coord-1.0.0.yaml
drdavella: I think we discussed this with regards to `Quantity`, and yes the documentation proposes that tag versions should move in lock-step with the overall ASDF version. However, it seems odd to label something brand new as anything other than 1.0.0, so maybe we can discuss this when we meet.

schemas/stsci.edu/asdf/wcs/step-1.2.0.yaml
drdavella: Yes, that makes sense.

schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
nden: I believe you need to use the `ndarray` tag.

schemas/stsci.edu/asdf/wcs/celestial_frame-1.1.0.yaml
nden: Ah, OK. I missed that.

schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
drdavella: For some reason implementing this is trickier than I expected, which is why I haven't been able to merge yet.

PR #143, opened 2017-07-06
Quantity update

drdavella:
Add support for `ndarray` values in `Quantity` objects

### DIFF
--- a/schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
+++ b/schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
@@ -21,6 +21,7 @@ properties:
         items:
           type: number
         minItems: 1
+      - $ref: "../core/ndarray-1.0.0"
   unit:
     description: |
       The unit corresponding to the values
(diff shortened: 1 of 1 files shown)


schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
```diff
@@ -21,6 +18,7 @@ properties:
         items:
           type: number
```
nden: Perhaps nested lists should be allowed using `anyOf`.

nden:
It would be good to have an example section as well. These get in the documentation and I think can be helpful.

schemas/stsci.edu/asdf/unit/quantity-1.1.0.yaml
drdavella: I believe it's the case that nested lists will be handled by `ndarray` since this is the type that `astropy.Quantity` uses to represent non-scalar values anyway. I think I would argue that if you include a literal nested array in a manually edited ASDF file, it should really be an `ndarray` anyway. Does that make sense? I don't think ASDF currently provides support for arbitrarily nested arrays outside of `ndarray`.

nden: Yes, you are right. The list will be converted to ndarray by `Quantity`. Then is the `array` option necessary at all?

nden:
LGTM 👍

PR #146, opened 2017-08-06
add tabular to versionmap

nden:
Schema for tabular model is missing from the version 1.1.0 map.


nden:
@drdavella Can we merge this - I am getting Warnings because `tabular` is not in versionmap 1.1.0.

PR #147, opened 2017-09-27
Update Time schema to support EarthLocation

drdavella:
The older `v1.0.0` schema for `astropy.Time` did not properly account for the `location` attribute. This PR adds a new `v1.1.0` schema that allows this property to be represented correctly.

### DIFF
--- a/schemas/stsci.edu/asdf/time/time-1.1.0.yaml
+++ b/schemas/stsci.edu/asdf/time/time-1.1.0.yaml
@@ -248,15 +248,14 @@ allOf:
               - `lat`: in degrees
               - `h`: optional height
 
-          anyOf:
-            - type: object
-              properties:
-                x:
-                  $ref: "../unit/quantity-1.1.0"
-                y:
-                  $ref: "../unit/quantity-1.1.0"
-                z:
-                  $ref: "../unit/quantity-1.1.0"
-              required: [x, y, z]
+          type: object
+          properties:
+            x:
+              $ref: "../unit/quantity-1.1.0"
+            y:
+              $ref: "../unit/quantity-1.1.0"
+            z:
+              $ref: "../unit/quantity-1.1.0"
+          required: [x, y, z]
 
       required: [value]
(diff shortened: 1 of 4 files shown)


schemas/stsci.edu/asdf/time/time-1.1.0.yaml
```diff
@@ -0,0 +1,262 @@
...
+          description: |
+            Specifies the observer location for scales that are
+            sensitive to observer location, currently only `tdb`.  May
+            be specified either with geocentric coordinates (X, Y, Z)
+            with an optional unit or geodetic coordinates:
+              - `long`: longitude in degrees
+              - `lat`: in degrees
+              - `h`: optional height
+
+          anyOf:
```
nden: Is this `anyOf` necessary here, given that there's only one option. Or am I misreading it?

schemas/stsci.edu/asdf/time/time-1.1.0.yaml
```diff
@@ -0,0 +1,262 @@
...
+            - type: object
+              properties:
+                x:
+                  $ref: "../unit/quantity-1.1.0"
+                y:
+                  $ref: "../unit/quantity-1.1.0"
+                z:
+                  $ref: "../unit/quantity-1.1.0"
+              required: [x, y, z]
```
nden: Are numbers + units not allowed any more? That's fine, I just want to confirm this was the intention.

schemas/stsci.edu/asdf/time/time-1.1.0.yaml
drdavella: Good catch, I think this was just an oversight.

schemas/stsci.edu/asdf/time/time-1.1.0.yaml
drdavella: No, that is no longer allowed, although v1.0.0 of the schema should still be interpreted properly.

drdavella:
@nden that occurred to me as well but I just wanted to do the simple fix first. I'll open an issue in `asdf`.

PR #149, opened 2017-11-28
Add schema for the externalarray type

Cadair:


nden:
We try to include a `description` field in all schemas. Could you please add one.
Also, when possible, include examples. Here's an example

https://github.com/spacetelescope/asdf-standard/blob/master/schemas/stsci.edu/asdf/unit/unit-1.0.0.yaml

schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - type: string
+  datatype:
+    type: string
+  shape:
+    type: array
+    items:
+      anyOf:
+      - type: integer
+        minimum: 0
+      - enum: ['*']
```
drdavella: It's not clear to me what purpose this serves.

schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
```diff
@@ -0,0 +1,25 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/core/externalarray-1.0.0"
+tag: "tag:stsci.edu:asdf/core/externalarray-1.0.0"
+title: Point to an array-like object in an external file.
+
```
drdavella: An example would be useful, and also a description, as @nden mentioned.

schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
Cadair: 🤷‍ I have no idea either. I will remove it and see if it breaks.

drdavella:
@Cadair https://github.com/spacetelescope/asdf/pull/400 seems ready for merge, but this schema could still use a brief description.

Cadair:
🎉 thanks!

PR #153, opened 2018-01-25
Reflect recent update to ICRS tag in Astropy

drdavella:
This was changed in astropy/astropy#7113 and is being tracked in https://github.com/spacetelescope/asdf/issues/431.


drdavella:
This will be overridden by #151, but it's necessary for ASDF tests to pass in the meantime.

PR #154, opened 2018-02-05
Fix erroneous use of "requiredProperties" in schemas

drdavella:
Several of the schemas were using `requiredProperties` instead of `required` to indicate schema properties that are required. This has now been fixed.

The use of the invalid property `requiredProperties` meant that these schemas were not being properly validated, which hid a few relatively minor bugs. These bugs have now been fixed.

@nden this should have no impact on the JWST pipeline, so I think this can be merged as soon as tests pass.

### DIFF
--- a/schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
+++ b/schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
@@ -20,7 +20,7 @@ examples:
 
 type: object
 properties:
-  source:
+  fileuri:
     type: string
   target:
     anyOf:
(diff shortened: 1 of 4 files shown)


schemas/stsci.edu/asdf/core/externalarray-1.0.0.yaml
```diff
@@ -20,7 +20,7 @@ examples:
 
 type: object
 properties:
-  source:
+  fileuri:
```
nden: Does this require a corresponding change in asdf?

drdavella: No, it was correct in ASDF but since required properties weren't being validated correctly, and since `additionalProperties` is set to `true`, the error in the schema wasn't picked up.

PR #155, opened 2018-02-05
Update top-level ASDF core schema to v1.1.0

drdavella:
This PR removes restrictions on the top-level `data`, `wcs`, and `fits` attributes.


nden:
So would this require all schema versions to be increased too?

drdavella:
My personal opinion is no, since no other schemas make reference to this one.

drdavella:
Good catch! I fixed it and will merge when CI passes here and in ASDF.

PR #159, opened 2018-04-09
Add a Linear1D schema

Cadair:


nden:
@drdavella  Since this is a new addition should it be added to version `1.1.0` and not `1.0.0`?
And also added to the corresponding version_map file?

drdavella:
@nden I go back and forth about this, but the convention we have been following is to start all new types/schemas at `1.0.0`. It makes some things more complicated, but to me it seems a little more intuitive than having to bump **all** tag and schema versions every time a change is made to any one of them.

PR #160, opened 2018-04-12
Add a schema for the new Multiply model

Cadair:
see astropy/astropy#7210


drdavella:
See discussion in https://github.com/astropy/astropy/pull/7210#discussion_r179283830 for motivation.

PR #166, opened 2018-06-15
Update published references

mwtoews:
Found a few old "in press" references that have been published.

For README.md, use blockquote style (rather than code block).


drdavella:
Thanks @mwtoews!

PR #170, opened 2018-07-30
Update reference files

drdavella:
I reran the `generate` script in `reference_files/generate`. It seems like this hasn't been done in a while.


drdavella:
We do not want to merge these reference examples with history sections that refer to a specific implementation/version of ASDF.

PR #177, opened 2018-10-24
This document **does** represent a released version of the Standard

drdavella:
I think this ship has sailed, whether we're ready or not.


perrygreenfield:
I won't be able to look at this until I'm finished with freeze work

PR #179, opened 2018-11-04
remove wcs schemas

nden:
See spacetelescope/asdf#554


drdavella:
@nden I'm wondering if instead of removing these wholesale, maybe the right approach is simply to omit them from the v1.3.0 version map but leave them in place.

nden:
Agreed, I made another PR but got side tracked with another issue.

drdavella:
@nden I'm closing this in favor of #182.

PR #181, opened 2018-11-05
Remove domain field from transform schema

drdavella:
This seems to me more of a bug fix since `domain` wasn't even included in version 1.2.0 of the standard. If this is the case, we shouldn't actually need to bump the version number of `transform`, which means we can avoid a cascade of changes throughout the rest of the transform schemas.

### DIFF
--- a/schemas/stsci.edu/asdf/transform/add-1.2.0.yaml
+++ b/schemas/stsci.edu/asdf/transform/add-1.2.0.yaml
@@ -1,0 +1,34 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/add-1.2.0"
+tag: "tag:stsci.edu:asdf/transform/add-1.2.0"
+title: >
+  Perform a list of subtransforms in parallel and then
+  add their results together.
+
+description: |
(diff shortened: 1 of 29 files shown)


nden:
`domain` was not in versin `1.2.0` because it was deprecated. The deprecation message was issued by the `Transform` tag. Essentially if an asdf object had `domain` the deprecation warning was issued and the next time the file was written, `domain` was converted to `bounding_box`. So yes, perhaps the version of the transform schema should be increased.

nden:
`TransformType` in astropy points to version `1.1.0`. This should be bumped as well.

drdavella:
@nden this will actually require lots of version bumps in Astropy since we'll have to create a lot of new schema versions in order to reflect the update to `transform`.

drdavella:
🤷‍♂️ 

It seems like we need to update the `transform` schema regardless so I'm not sure it would make things that much easier.

It does make me wonder whether all of the transform schemas actually belong in Astropy as well, but I'm not too enthusiastic about making that change. Right now the only thing that would ever motivate a change to these schemas is a change in Astropy, but maybe that's not true in the long term.

drdavella:
Ironically these tests won't pass until we merge the associated PR in ASDF, so I think we'll just go ahead and merge here. It passes for me locally.

PR #183, opened 2018-11-15
Prevent conftest from being used as submodule

drdavella:
This prevents a warning from occurring with newer versions of `pytest` when running the test suite of the ASDF package.


drdavella:
Merging in spite of unrelated test failures.

PR #184, opened 2018-12-08
add schema for unit equivalencies

nden:
This is to be merged only if there's an agreement and astropy/astropy#8252 is merged.
May need to change the version.


drdavella:
I think we'll need to bump to v1.4.0 for this since we've already released an `asdf` with v1.3.0 of the standard.

nden:
@drdavella Will do. Let's see what happens with the other PR.

nden:
@drdavella It looks like one of the error is a bug in `convert_schema`. Not sure about the rest of the errors. Any ideas?

drdavella:
The other failures are unrelated and I'll need to account for them elsewhere. I'll look into the problem with `convert_schemas`.

nden:
sigh :-(

drdavella:
If you rebase it should fix most of the travis failures. The tests of the examples in the new schema will continue to fail though since there's not an implementation available yet.

nden:
Sorry @drdavella, I overwrote your changes by mistake.

drdavella:
@nden I'm sorry for just getting back to this now. I wonder if this schema actually belongs in `astropy` since it seems fairly specific to the way that unit equivalencies work there? It also might make a bit more sense logically, but I can be persuaded otherwise.

nden:
Moved the schema to astropy. Closing this.

PR #199, opened 2019-04-18
Improve schema documentation

### LINKED ISSUE #145: Add documentation for extending an existing schema
This is currently missing on [the documentation page](http://asdf-standard.readthedocs.io/en/latest/extending.html#extending-an-existing-schema).

drdavella:
This PR adds substantial content to the documentation about schemas. It also makes some improvements to the example for creating new schemas and tags. It adds some documentation about extending existing schemas (and therefore closes #145). Finally, it bumps some schema versions that were previously overlooked and updates the references to the new schema versions in the documentation.

@perrygreenfield if you had a few minutes to read just the documentation component of this and provide feedback, that would be great.


drdavella:
I'm going to go ahead and merge so the documentation is live, which should make it easier to provide feedback.

drdavella:
See https://asdf-standard.readthedocs.io/en/latest/schemas.html for the live docs.

PR #203, opened 2019-08-08
Add orthogonal polynomial model support

perrygreenfield:
Previously only ordinary polynomials were supported in ASDF. This allows astropy orthogonal polynomial models to be serialized.


perrygreenfield:
messed up, will issue new PR

PR #206, opened 2019-09-12
add fix_inputs schema

perrygreenfield:
This adds the schema to support the fix_inputs operation in modeling compound models.


nden:
This is not a valid schema. I'm going to close it and create a new one (also because it's not based on master).

PR #212, opened 2019-10-25
add grating equation schema

nden:
Adds a schema for a grating equation model. Two models in astropy will use it: `WavelengthFromGratingEquation` and `AnglesFromGratingEquation`.


nden:
Moved to gwcs.

PR #214, opened 2020-02-06
Add YAML end-of-document marker to schema files

eslavich:
This change isn't particularly important, but I think it'll be helpful for the schemas to have some uniformity in structure.  I'm submitting this PR separately so that it doesn't clutter up the review of the main cleanup PR, but there will be a test to confirm that `%YAML 1.1\n---` and `\n...\n` are present at the end of each schema file.


jdavies-st:
Good to have uniformity.  👍

jdavies-st:
Will be happy to see the test that checks these, and maybe include them in those packages, or have them be a part of the schema checker provided by `asdf`, in addition to being unit tests in this package.

eslavich:
Including this kind of thing in the asdf schema checker is a good idea, though we wouldn't want to require it -- YAML documents are still valid without the `...`.  Maybe add a config option `asdf_schema_strict`?

jdavies-st:
That seems reasonable.

PR #216, opened 2020-02-06
Restructure repo and revive Travis build

eslavich:
This PR makes the following changes:

- Add more complete .gitignore
- Fix .travis.xml and convert to tox jobs
- Add docs-requirements.txt and test-requirements.txt
- Drop support for older asdf versions from conftest.py
- Move documentation root from repo root to docs/ and update conf.py
- Create tox config with schema validation, black, flake8, and doc build checks


.gitignore
```diff
@@ -1,5 +1,132 @@
...
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+# Ignore generated Sphinx documentation
+docs/source/generated/
```
jdavies-st: Need "\n"

nden:
My only comment is that the asdf-standard pointer in asdf should be updated after each of these large structural PRs and tests in asdf, astropy, gwcs and jwst (if possible) run with the dev versions.

eslavich:
I ran the tests on each of those repos with the schemas in #217 (which includes all changes in this branch) and they all passed.

PR #217, opened 2020-02-06
Additional schema tests and resulting cleanup

eslavich:
This is the main course of my cleanup effort.  New tests include:

For each schema:
- Confirm that filename follows our convention thus far (name with lowercase letters, underscores, and numbers, followed by a 3-part version and .yaml extension)
- Confirm presence of YAML header and footer
- Confirm that `$schema` key exists and is set to yaml-schema/draft-01
- Confirm that the `id` key exists, matches the filename, and is unique among all the schemas
- If the `tag` key is present, confirm that it matches the filename, and is unique among all the schemas
- Confirm that the `title` key is present and has a value
- Confirm that the `description` key is present and has a value
- Confirm that examples reference the version of the schema described by the file (and not some earlier version, from a copy/paste)
- Confirm that refs in the description reference the version of the schema described by the file

For each transform schema:
- Confirm that schema includes another transform schema or base transform schema

For the latest version of each schema:
- Confirm that `$ref` values only reference other latest-version schemas
- Confirm that examples only include other latest-version schemas
- Confirm that refs in the description reference only other latest-version schemas

For each metaschema:
- Confirm presence of YAML header and footer
- Confirm valid YAML

For each version map:
- Confirm that filename follows our convention (version_map-x.y.z.yaml)
- Confirm presence of YAML header and footer
- Confirm that the only keys are FILE_FORMAT, YAML_VERSION, and tags
- Confirm that FILE_FORMAT and YAML_VERSION have recognized values (currently only 1.0.0/1.1)
- Confirm that all the tags listed exist
- Confirm that the tag list doesn't contain duplicate names
- Confirm that the tag list is sorted alphabetically by tag name
- Confirm that tags did not disappear between successive version maps (except those that were explicitly deprecated, like transform/domain)

For the latest version map:
- Confirm that tags include the latest version of every non-deprecated tag

These tests revealed many issues that are all resolved here.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,166 @@
+import pytest
+
+from common import (
+    SCHEMAS_PATH,
+    load_yaml,
+    assert_yaml_header_and_footer,
+    VALID_SCHEMA_FILENAME_RE,
+    split_id,
+    list_refs,
+    path_to_id,
+    path_to_tag,
+    ref_to_id,
+    list_schema_paths,
+    list_latest_schema_paths,
+    DEPRECATED_ID_BASES,
+    METASCHEMA_ID,
+    list_example_ids,
+    list_description_ids,
+)
+
+
+@pytest.fixture(scope="session")
+def schemas():
+    return [load_yaml(p) for p in list_schema_paths(SCHEMAS_PATH)]
+
+
+@pytest.fixture(scope="session")
+def latest_schemas():
+    return [load_yaml(p) for p in list_latest_schema_paths(SCHEMAS_PATH)]
+
+
+@pytest.fixture(scope="session")
+def latest_schema_ids(latest_schemas):
+    result = set()
+    for schema in latest_schemas:
+        if "id" in schema:
+            result.add(schema["id"])
+    return result
+
+
+@pytest.fixture(scope="session")
+def latest_schema_tags(latest_schemas):
+    result = set()
+    for schema in latest_schemas:
+        if "tag" in schema:
+            result.add(schema["tag"])
+        elif "anyOf" in schema:
+            for elem in schema["anyOf"]:
+                if "tag" in elem:
+                    result.add(elem["tag"])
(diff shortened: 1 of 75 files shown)


eslavich:
@nden can you think of any checks that I'm missing?  Or any of the above that aren't actually useful/appropriate?

eslavich:
I installed asdf master with this branch of the schemas, and ran the unit tests on asdf, jwst, astropy, and gwcs.  Happy to report that everything passes!

eslavich:
All of the preceding PRs have been merged, so this is ready for review

tests/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+        if "id" in schema:
+            if schema["id"] not in result:
+                result[schema["id"]] = []
+            result[schema["id"]].append(schema)
+    return result
+
+
+@pytest.fixture(scope="session")
+def assert_schema_correct(tag_to_schema, id_to_schema):
+    def _assert_schema_correct(path):
```
jdavies-st: You may want to put a `__traceback__ = False` in this assertion helper so that when there is a failure, you get just the rewritten assertion, not the the traceback of the assertion helper.  This is a handy `pytest` feature.

tests/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+                if description_id_base == id_base and description_id != schema["id"]:
+                    assert False, f"{path.name} descriptioon contains an outdated ref"
+
+    return _assert_schema_correct
+
+
+@pytest.fixture(scope="session")
+def assert_latest_schema_correct(latest_schema_ids):
+    def _assert_latest_schema_correct(path):
+        schema = load_yaml(path)
```
jdavies-st: `__traceback__ = False` here too

tests/conftest.py
eslavich: Thanks, I'll do that

PR #221, opened 2020-02-11
Add subclass_metadata-1.0.0.yaml schema

eslavich:
The Python ASDF library references a core schema called subclass_metadata, but it doesn't exist yet.  This PR adds that missing schema.

Resolves https://github.com/spacetelescope/asdf/issues/748

### DIFF
--- a/schemas/stsci.edu/asdf/core/subclass_metadata-1.0.0.yaml
+++ b/schemas/stsci.edu/asdf/core/subclass_metadata-1.0.0.yaml
@@ -13,7 +13,8 @@ type: object
 properties:
   name:
     description: |
-      The name of the subclass to deserialize.
+      The name of the subclass that represents this object
+      when deserialized.
     type: string
 
 required: [name]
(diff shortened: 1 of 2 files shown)


schemas/stsci.edu/asdf/core/subclass_metadata-1.0.0.yaml
```diff
@@ -0,0 +1,20 @@
...
+description: |
+  Identifies the specific subclass that was serialized,
+  to enable ASDF readers to correctly deserialize the object.
+
+tag: "tag:stsci.edu:asdf/core/subclass_metadata-1.0.0"
+type: object
+properties:
+  name:
+    description: |
+      The name of the subclass to deserialize.
```
jdavies-st: In addition to deserialize, doesn't this also control serialization as well?

eslavich: Hmm... I'm not sure how to think about that.  I'll change the wording a bit.

eslavich:
I agree that we may want to deprecate this feature, as subclass names seem tied to a particular implementation of an ASDF library.  I'll raise that issue elsewhere, but for now I'd like to merge this just to get the current state of the code/schemas consistent with itself.

PR #224, opened 2020-02-19
Gaussian models serialization

shannnnnyyy:
added schemas for Gaussian1D and 2D models.

PR #225, opened 2020-03-04
Functional model schemas

shannnnnyyy:
Added schemas for all of the functional models except the few which are already done under polynomial (linear, multiply, shift, scale). If this is merged, #224 can be closed because this also includes the gaussians.


schemas/stsci.edu/asdf/transform/sersic1d-1.0.0.yaml
```diff
@@ -0,0 +1,33 @@
...
+        anyOf:
+          - $ref: "../unit/quantity-1.1.0"
+          - type: number
+        description: Surface brightness at r_eff.
+      r_eff:
+        anyOf:
+          - $ref: "../unit/quantity-1.1.0"
+          - type: number
+        description: Effective (half-light) radius.
+      n:
```
eslavich: @nden @perrygreenfield do you have an opinion on matching the modeling package parameter names vs using a name that is more descriptive?  If someone were to inspect an asdf file written under this schema, they'd just see `n` and the schema description wouldn't be readily available.  We could name it `sersic_index`, for example.

shannnnnyyy:
yep, I'll add tests and make a new version map. speaking of - i'm trying to figure out why the tests are failing for #224 and I thought it was because the new schemas weren't in the map so I added them but the tests are still failing. even though that can be closed once this PR is good to go, i'd like to get those tests passing there now so I know exactly what I need to do for this one.

schemas/stsci.edu/asdf/transform/sersic1d-1.0.0.yaml
perrygreenfield: I think it would be good practice for the ASDF representation to be more descriptive. Nothing requires it to use the same name that the modeling package uses.

shannnnnyyy: in this particular case i personally think the parameter n is widely understood to be the sersic index, but there are certainly other cases where this issue would apply

nden: I think names which are widely accepted should be used both in modeling and asdf.
For models which have definitions in the VO standard I think we should try to use the same names. I was going to suggest checking all names against the VO definitions when moving the transforms out of the standard.

eslavich:
> yep, I'll add tests and make a new version map. speaking of - i'm trying to figure out why the tests are failing for #224 and I thought it was because the new schemas weren't in the map so I added them but the tests are still failing. even though that can be closed once this PR is good to go, i'd like to get those tests passing there now so I know exactly what I need to do for this one.

This turned out to be an issue with the schema testing code -- it couldn't handle schemas whose Type classes don't exist yet.  I opened a PR (https://github.com/spacetelescope/asdf/pull/771) to fix the problem.

schemas/stsci.edu/asdf/transform/airydisk2d-1.0.0.yaml
```diff
@@ -0,0 +1,44 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/airydisk2d-1.0.0"
```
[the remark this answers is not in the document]
shannnnnyyy: good point, fixed this!

schemas/stsci.edu/asdf/transform/box1d-1.0.0.yaml
```diff
@@ -0,0 +1,43 @@
...
+description: >
+  One dimensional box.
+
+examples:
+  -
+    - A 1D box of width 4.0, amplitude 10.0, centered at x=1.5. 
+    - |
+      !transform/box1d-1.0.0
+        amplitude: 10.0
+        bounding_box: [-0.5, 3.5]
```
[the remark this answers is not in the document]
shannnnnyyy: should I go ahead make 1.3.0 and then have just the new models use this version? also, why is 1.1.0 the exact same as 1.0.0 (similarly, when does it make sense to make a new version of transform?)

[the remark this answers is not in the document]
eslavich: @nden what do you think?  I'm inclined to say yes, just so we don't have to release the standard again right away.

I don't know why 1.1.0 was created with no changes -- it could be that at the time we assumed that the schema versions had to stay identical to the standard version.

nden:
This looks good! One thing I noticed is a `transform` schema is not added to `versionmap-1.5.0`.

@eslavich Does it make sense to add a check in the tester that a version of every schema is included in versionmap files?

eslavich:
> This looks good! One thing I noticed is a `transform` schema is not added to `versionmap-1.5.0`.
> 
> @eslavich Does it make sense to add a check in the tester that a version of every schema is included in versionmap files?

There is a check, [here](https://github.com/spacetelescope/asdf-standard/blob/master/tests/test_version_map.py#L65), but I removed schema fragments (such as transform) from the version map in my refactor, since they don't have a `tag` field and aren't intended to be directly serializable.

shannnnnyyy:
In my latest commit I added some new non-functional-model schemas in here too. 

@nden by `a transform schema is not added to versionmap-1.5.0`, do you mean that transform/transform-1.x.x all be in there too? If thats the case, that's not happening in the old version maps either. Also, should I create 1.3.0 and include the bounding box?

nden:
@shannnnnyyy Ed clarified he deliberately removed schema fragments, like `transform-1...` from versionmap files. 

I am not sure what the best way to deal with `bounding_box` is. It is currently handled in the tag code as an optional attribute and is validated in modeling at some level.  Does it belong in the standard or is it a Python specific implementation detail? Also is it going to change in the future? Perhaps this needs more discussion.

shannnnnyyy:
oh sorry i missed that comment! i'll leave the 1.3.0 thing alone then, and just keep adding new schemas here as I write them.

schemas/stsci.edu/asdf/version_map-1.5.0.yaml
```diff
@@ -36,6 +37,7 @@ tags:
   tag:stsci.edu:asdf/transform/cylindrical_perspective: 1.3.0
   tag:stsci.edu:asdf/transform/disk2d: 1.0.0
   tag:stsci.edu:asdf/transform/divide: 1.2.0
+  tag:stsci.edu:asdf/transform/druded: 1.0.0
```
nden: typo -> `drude1d`

PR #226, opened 2020-03-05
Ignore unrecognized tag warnings in schema tests in this repo

eslavich:
This PR fixes the default resolver override in conftest.py that was broken by changes to resolver code in asdf.  It also disables warnings on unrecognized objects when deserializing schema examples in tests.

These changes will allow the tests to function when new schemas are added to this repo.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,19 +1,29 @@
 import os
 
 import asdf
-from asdf import resolver as res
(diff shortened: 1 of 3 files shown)


eslavich:
Doing another rush merge here, but I'll follow up on any comments with an additional PR.

PR #228, opened 2020-04-06
more schemas

shannnnnyyy:
I added a schema for const2d, and for each of the powerlaw models. I think this finishes up serializing all of the astropy models - everything in functional models, physical models, powerlaws is done in addition to tabular, projections, and rotations.

(note: instead of just using the existing 'constant-1.x.x' schema I had to create a new 'constant2d-1.0.0' schema. Because there is nothing in the schema that determines if it should be a Const1D or Const2D model (it only contains 'value'), I had to create a unique tag instead of handling both in the existing ConstantType and a unique schema so both don't point to the same one.)


eslavich:
> (note: instead of just using the existing 'constant-1.x.x' schema I had to create a new 'constant2d-1.0.0' schema. Because there is nothing in the schema that determines if it should be a Const1D or Const2D model (it only contains 'value'), I had to create a unique tag instead of handling both in the existing ConstantType and a unique schema so both don't point to the same one.)

Another option is to add a new field to the existing schema, `dimensions` or the like, and have the serializer build the model class according to that field's value (defaulting to 1D).  I don't know if that's better in any way -- I suppose it would easily generalize to 3D, 4D, etc if needed, but I don't know how useful that is.

schemas/stsci.edu/asdf/transform/brokenpowerlaw1d-1.0.0.yaml
```diff
@@ -0,0 +1,44 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/brokenpowerlaw1d-1.0.0"
```
eslavich, line 4: Shall we continue the proud tradition of snake case filenames and ids/tags?

shannnnnyyy: whoops forgot about that!

schemas/stsci.edu/asdf/transform/constant2d-1.0.0.yaml
```diff
@@ -0,0 +1,21 @@
...
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/constant2d-1.0.0"
+tag: "tag:stsci.edu:asdf/transform/constant2d-1.0.0"
+title: >
+  A transform that takes no inputs and always outputs a constant
+  value.
+description: |
+  Invertibility: All ASDF tools are required to be able to compute the
+  analytic inverse of this transform, which always outputs zero values.
```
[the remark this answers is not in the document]
shannnnnyyy: yeah i did, didn't catch that thanks

schemas/stsci.edu/asdf/transform/constant-1.4.0.yaml
```diff
@@ -0,0 +1,23 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/constant-1.4.0"
+tag: "tag:stsci.edu:asdf/transform/constant-1.4.0"
+title: >
+  A transform that takes no inputs and always outputs a constant
+  value.
```
nden, line 8: This is not quite right because it takes 1 or 2 inputs based on the dimensionality. Perhaps just say
"A Constant transform".
And add more to the description.

schemas/stsci.edu/asdf/transform/constant-1.4.0.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: |
+  Invertibility: All ASDF tools are required to be able to compute the
+  analytic inverse of this transform, which always outputs zero values.
```
nden, line 11: Sorry, I don't understand this description. In general Const1(2)D models don't have an analytical inverse.I would suggest
"A transform which takes one or two inputs based on dimensionality and returns a constant value. It has no analytical inverse.

schemas/stsci.edu/asdf/transform/constant-1.4.0.yaml
```diff
@@ -0,0 +1,23 @@
...
+allOf:
+  - $ref: "transform-1.2.0"
+  - type: object
+    properties:
+      value:
+        anyOf:
+          - $ref: "../unit/quantity-1.1.0"
+          - type: number
+      dimensions:
+        type: number
```
nden, line 21: The type should be "integer".

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+  - If a constant, it must be a single item which is a constant value
+    to send to the output axis.
+
+  If only a list is provided, the number of input axes is
+  automatically determined from the maximum index in the list.  If an
+  object with `mapping` and `n_inputs` properties is provided, the
+  number of input axes is explicitly set by the `n_inputs` value.
+
+  Invertibility: TBD
```
nden, line 27: Suggested text for Invertibility:
"This transform does not have a general analytical inverse. In some well defined cases it is possible to invert automatically."

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
```diff
@@ -0,0 +1,94 @@
...
+          mapping: [0, 1]
+          n_inputs: 3
+
+definitions:
+  mapping:
+    type: array
+    items:
+      anyOf:
+        - type: integer
+        - $ref: "../core/constant-1.0.0"
```
nden, line 80: I wonder why this is defined as a constant and not integer. If I had to write the schema I would have written the items as integers. Any ideas?

eslavich: I can't figure out why we have core/constant at all -- it's only used in remap_axes, and it doesn't seem like the RemapAxesType in astropy can ever write a core/constant:

https://github.com/astropy/astropy/blob/master/astropy/io/misc/asdf/tags/transform/compound.py#L143

And on read, it turns a non-integer node into a Const1D anyway.  Maybe core/constant is a failed experiment that just hasn't been cleaned up yet?

schemas/stsci.edu/asdf/transform/label_mapper-1.3.0.yaml
```diff
@@ -0,0 +1,149 @@
+%YAML 1.1
```
eslavich, line 1: We don't need to add this file anymore, right?

schemas/stsci.edu/asdf/transform/regions_selector-1.3.0.yaml
```diff
@@ -0,0 +1,99 @@
+%YAML 1.1
```
eslavich, line 1: Nor this one...

schemas/stsci.edu/asdf/transform/label_mapper-1.3.0.yaml
shannnnnyyy: I wasn't sure if these still required a new version since they are linked in an updated example so I kept them, but I'll remove them now and see if the tests pass

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
eslavich: Per discussion on Slack, let's remove core/constant here, since the astropy [tag class](https://github.com/astropy/astropy/blob/master/astropy/io/misc/asdf/tags/transform/compound.py#L148) no longer writes it anyway.

shannnnnyyy: should i remove core/constant from the newest version map too?

nden:
I wonder if all schemas in wcs/ need to be removed from version_map-1.5.0. I tested jwst without those schemas. Or do we have to carry them forever, in case there are some really old files that use them. Though it's unlikely there are such old files, the schemas were moved to gwcs ~ 2.5 years ago.

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
eslavich: I think we should leave it for now, the value of that schema isn't clear to me but it's been in the core schemas for a while.  We should probably seek community input before removing it.

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
```diff
@@ -0,0 +1,91 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/asdf/transform/remap_axes-1.3.0"
+tag: "tag:stsci.edu:asdf/transform/remap_axes-1.3.0"
+title: >
+  Reorder, add and drop axes.
+
+description: |
```
eslavich, line 9: This description needs to be updated, now that the option to use a constant has been removed.

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
```diff
@@ -0,0 +1,91 @@
...
+          mapping: [1]
+
+  -
+    - For 2 input axes, return the first axis twice, followed by the second
+    - |
+        !transform/remap_axes-1.3.0
+          mapping: [0, 0, 1]
+
+  -
+    - For 2 input axes, add a third axis which is a constant
```
eslavich, line 49: I think we can do away with this example, it doesn't seem distinct from the previous one now that the constant has been removed.

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
```diff
@@ -0,0 +1,91 @@
...
+    - |
+        !transform/remap_axes-1.3.0
+          mapping: [0, 1, 42]
+
+  -
+    - |
```
eslavich, line 55: Is it still valuable to show an example of fixing a constant?  If so, we should keep the transform/constant that was removed (and change the description, since this example is no longer equivalent to one above).  If not, this example can be removed.

eslavich:
> I wonder if all schemas in wcs/ need to be removed from version_map-1.5.0. I tested jwst without those schemas. Or do we have to carry them forever, in case there are some really old files that use them. Though it's unlikely there are such old files, the schemas were moved to gwcs ~ 2.5 years ago.

I think that would be fine, we do have to keep the individual schemas for as long as we want to support older ASDF Standard versions, but they have no purpose in the latest version map if gwcs is already using its own schemas.

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
shannnnnyyy: are you referring to the example that starts with 'For 2 input axes, return the first axis twice, followed by the second' or the one below that?

eslavich: Those are the two that I think (could be wrong) are the same.  I would get rid of the one with `42` in it, since a model with 43 inputs seems unlikely...

schemas/stsci.edu/asdf/transform/remap_axes-1.3.0.yaml
nden: Yes, this needs to be removed

eslavich:
There are some warnings during the docs build:

```
blackbody-1.0.0.rst:1: WARNING: Inline interpreted text or phrase reference start-string without end-string.
rotate_sequence_3d-1.0.0.rst: WARNING: Could not lex literal_block as "yaml". Highlighting skipped.
```

The first can be fixed in the blackbody schema by changing the description field to use `|` instead of `>`.

The second can be fixed in the rotate_sequence_3d schema by removing the trailing tab character from the line `rotation_type: cartesian`.

eslavich:
I built the docs and cruised through all the transform schemas, and the only rendering issue I found besides the two due to the warnings I mentioned above is in broken_power_law1d.  The existing example description is this:

```
$$f(x) = A (10.0 / 5.0) ^ {-2.0} for x < 5.0$$ and $$ f(x) = A (10.0 / 5.0) ^ {-3.0} for x > 5.0$$
```

If I change to this, it seems to work:
```
$f(x) = A (10.0 / 5.0) ^ {-2.0}$ for x < 5.0 and $f(x) = A (10.0 / 5.0) ^ {-3.0}$ for x > 5.0
```

PR #258, opened 2020-06-22
Add schema collections

eslavich:
This adds a schema that describes the schema collection document, and also adds a schema collection for the schemas that will be part of the standard in 2.0.0.

This will probably change, but at least it'll get something into the epic branch for us to mess around with in the transforms repo.


perrygreenfield:
Looks straightforward enough, though I haven't yet reviewed the directories to see if anything has been missed. I didn't see the gargleblaster schema though.

eslavich:
I'm still puzzling over what to do about the dependencies.  These lines:

```
dependency_ids:
  - http://asdf-format.org/schema_collections/core-2.0.0
```

show that schemas in this collection reference core schemas in the core-2.0.0 collection, which will let us show helpful warning messages when core-2.0.0 is not installed.  But if we make a backwards-compatible change to the core schemas (say we add a new schema but don't change anything else) and release core-2.0.1, do we really want to have to go in and create new collections everywhere core is referenced?

Maybe instead these dependency requirements ought to support a range:

```
requirements:
  http://asdf-format.org/schema_collections/core: ~= 2.0.0
```

Now we can release as many 2.0.x versions of core as we want, and we won't have to touch the transforms.

perrygreenfield:
Wouldn't you want to only allow minor changes in the standard (and that would imply that those changes were backward compatible), or also intermediate changes, but not major version ones. Since the file can't see the future, the restrictions will have to be on the future by some system, right?

eslavich:
I think that would work?  If we're responsible about obeying some flavor of semantic versioning, then all necessary attributes should be present and an ASDF reader wouldn't have any trouble ignoring the unrecognized stuff.

This pinned vs range issue also rears its head when a schema references another:

```
foo:
  tag: http://stsci.edu/schemas/asdf/core/ndarray-1.0.0
```
This schema wants to validate that there is an ndarray in the `foo` field.  Presumably in most cases schema authors will not care if an ndarray-1.0.1 is also accepted?

Maybe in both cases we ought to permit a range.  Not sure if it's better to have the schema/collection author make the range explicit, or to have the software interpret any version as an implicit range.

eslavich:
If we don't want to implement a full Pythonesque version specification, we could just allow a simple wildcard:

```
http://stsci.edu/schemas/asdf/core/ndarray-1.0.*
```

perrygreenfield:
The problem I see with range is that you know there is an end version that isn't compatible (otherwise, what use is it?). And to know that means you have to issue a new schema to make that change, and if you do that, you are doing what you are trying to avoid, aren't you? What use case would that be useful for?

eslavich:
> The problem I see with range is that you know there is an end version that isn't compatible (otherwise, what use is it?). And to know that means you have to issue a new schema to make that change, and if you do that, you are doing what you are trying to avoid, aren't you? What use case would that be useful for?

Ohhh right.  So this isn't like software dependencies because the schema is immutable once released.  I agree that a range wouldn't be useful.

I _think_ there's still some utility in being able to select the version number that's flexible.  This would be the wildcard option.  Only because it's clear that not every developer can be trusted to obey strict semantic versioning, and for schemas from some sources users may wish to tighten the version number.

eslavich:
Oops, actually, I think it's still important to be able to specify the minimum part of the range.  If a feature was added in 1.5 that we need, then there's no way to express `>= 1.5, < 2.0` using just wildcards.

eslavich:
This has been superseded by #272

PR #272, opened 2020-08-06
Add extension manifest schema

eslavich:
This is the schema that validates an extension manifest document, which will list tags associated with a specific version of an extension to ASDF.  Note several changes to the URI structure (which are all up for discussion):

Old URI style: `http://stsci.edu/schemas/asdf/core/extension_manifest-1.0.0`
New URI style: `asdf://asdf-format.org/core/schemas/extension_manifest-1.0`

- Change scheme from `http://` to `asdf://`.  Users report that `http://` is confusing because it seems to imply that the resource is available from that web address, or that the library is downloading something.
- Change authority from `stsci.edu` to `asdf-format.org`.
- Change path from `schemas/asdf/core` to `core/schemas`.  With the `asdf` scheme, there's no need to leave room for non-asdf schemas.  I swapped `core` and `schemas` because we're adding several new entity types (extensions, extension manifests, tags, binary transforms, etc) and I think it'll be more convenient to keep all core entities together in one namespace instead of distributed among several.  Tentatively I'm planning to give each official extension a dedicated path under `extensions`, e.g., `asdf://asdf-format.org/extensions/transform/schemas/...`.
- Drop the patch version.  To date we've never used a nonzero patch version on any of the schemas or tags.  I can't think of a scenario where it would be useful to do so.

### DIFF
--- a/resources/asdf-format.org/core/schemas/extension_manifest-1.0.yaml
+++ b/resources/asdf-format.org/core/schemas/extension_manifest-1.0.yaml
@@ -38,7 +38,7 @@ properties:
       Long description of the extension.
     type: string
 
-  asdf_standard_version:
+  asdf_standard_requirement:
     description: >
       ASDF Standard version requirement.
     anyOf:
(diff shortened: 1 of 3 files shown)


eslavich:
> Do we actually have an extension manifest that comports with this schema?

Here's a sample with a subset of the transform schemas: https://gist.github.com/eslavich/2fe24ad14cb11e71935a555431c519ac

PR #276, opened 2020-10-13
Update documentation for 1.6.0 changes

eslavich:
Included are new sections on:

- Handling the `default` schema annotation
- Restricted YAML features
- Handling null values in the tree

### DIFF
--- a/docs/source/file_layout.rst
+++ b/docs/source/file_layout.rst
@@ -83,8 +83,9 @@ Each line must begin with a ``#`` character.
 Tree
 ----
 
-The tree stores structured information using `YAML Ain’t Markup
-Language (YAML™) 1.1 <http://yaml.org/spec/1.1/>`__ syntax.  While it
+The tree stores structured information using a subset of `YAML Ain’t Markup
+Language (YAML™) 1.1 <http://yaml.org/spec/1.1/>`__ syntax (see :ref:`yaml_subset` for
+details on YAML features that are excluded from ASDF).  While it
 is the main part of most ASDF files, it is entirely optional, and a
 ASDF file may skip it completely.  This is useful for creating files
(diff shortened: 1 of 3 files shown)


docs/source/schemas.rst
```diff
@@ -475,6 +475,30 @@ The ``allOf`` combiner means that any instance that is validated against
 ``software_extended-1.0.0`` will have to conform to both the base software schema
 and the properties specific to the extended schema.
 
+Default annotation
+------------------
+
+The JSON Schema spec includes a schema annotation attribute called ``default`` that
```
perrygreenfield, line 481: Might be worth a very brief comment on why this changed, and also in that JSONSchema doesn't have well defined semantics for default either.

perrygreenfield (commented):
Looks great. I think it only needs one small change.

docs/source/file_layout.rst
```diff
@@ -83,8 +83,9 @@ Each line must begin with a ``#`` character.
 Tree
 ----
 
-The tree stores structured information using `YAML Ain’t Markup
-Language (YAML™) 1.1 <http://yaml.org/spec/1.1/>`__ syntax.  While it
+The tree stores structured information using a subset of `YAML Ain’t Markup
+Language (YAML™) 1.1 <http://yaml.org/spec/1.1/>`__ syntax (see :ref:`yaml_subset` for
+details on YAML features that are excluded from ASDF).  While it
 is the main part of most ASDF files, it is entirely optional, and a
```
kmacdonald-stsci, line 89: Should this be "an ASDF" or "a ASDF"?

eslavich: Hmm... I don't think of ASDF as a countable noun, since the F stands for Format and not File.  How about "excluded from ASDF files"?

PR #277, opened 2021-01-11
Document change for removal of 52-bit limit for string literals of numbers

kmacdonald-stsci:
Within the tree, it is possible to represent numeric values as string literals in the ASDF file.  For practical reasons, there is a limit to the size of these numbers.  Originally, this limit reflected the 52-bit limit of javascript, with the anticipation of other languages interpreting ASDF files.  This limit has been deemed unnecessary, so has been removed.


docs/source/tree.rst
```diff
@@ -223,13 +223,11 @@ allowed.
...
-While it is possible to store arbitrary-sized integers as literals in
-YAML, not all programming languages and YAML libraries are able to
-read them.  Therefore, to ensure portability, all numeric literals in
-the tree must assume that the reader has no more precision than that
-of a 64-bit double precision floating point number: 52-bits of
-precision.  Therefore, ASDF libraries should refuse to write files
-containing integers that are larger than 52-bits.
+To store arbitrary-sized integers as literals in YAML and for YAML
+libraries, and the ASDF API, to properly read and interpret them,
+integers represented as string literals in the tree must be no more
```
eslavich, line 228: This:
```
integers represented as string literals in the tree must be no more than 64-bits.
```
seems redundant with the last sentence:
```
For integers, the string literal must be within the range represented by int64 types.
```
```

File: tox.ini [hunk 2]
```diff
@@ -9,22 +9,22 @@ commands =
     pytest
 
 [testenv:black]
-basepython = python3.8
+basepython = python3.9
```
eslavich, line 12: @jdavies-st recently pointed out to me setting`basepython` here has a downside, which is that we can't do this:

```
tox -e py38-black
```

We've been removing `basepython` in other repos, can you do that here too?

eslavich (commented):
Just a couple of little things, thanks for setting up actions too!

docs/source/tree.rst
jdavies-st: Instead of having a compound dependent clause, perhaps something like this is clearer:

> Integers represented as string literals in the ASDF tree must be no more than 64-bits. This limit exists because the YAML libraries used to interpret them must be able to round-trip the values exactly.  This is to ensure portability.

jdavies-st (approved):
Looks good! Just a small comment about the style in the documentation.

docs/source/tree.rst
kmacdonald-stsci: Qualifying 64-bits with `int64` isn't redundant because 64-bit integers can be interpreted in two different ways, `uint64`, and `int64`.  Also, I didn't run into any problems with YAML parsing for very large numbers in generic arrays, but arose when putting them in `ndarrays` in `numpy`, which are valid for only specific `dtypes`.

eslavich: Oh, I see, does "int64" signify "signed 64-bit integer"?  Is that a standard term of some kind?  I'd be inclined to spell out the "signed" so dummies like me don't get confused.

kmacdonald-stsci: Yes.  An unsigned 64-bit integer is termed `uint64`, whereas a signed 64-bit integer is termed `int64`.  In the documents it might be good to say "signed 64-bit integer" as well as `int64`.

jdavies-st (approved):

PR #284, opened 2021-05-03
Fix metaschemas so that custom validators on nested objects can be checked

### LINKED ISSUE #283: YAML and ASDF metaschemas do not validate beyond top level object
The YAML and ASDF metaschemas validate properties like `ndim`, `datatype`, and `tag`, but only do so when those properties are located at the top level object of a schema.  Occurrences of `ndim` and friends that are nested within other properties are ignored.

This can be fixed by adding in some of the material from https://json-schema.org/draft-04/schema.

eslavich:
Until now the metaschemas have been unable to check custom validators present in nested objects in the schema.  This PR redefines relevant JSON schema validators in terms of the current metaschema document to fix that issue.

Resolves #283

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -250,7 +250,7 @@ asdf_schema_standard_prefix = 'stsci.edu/asdf'
 # plugin.
 asdf_schema_reference_mappings = [
     ('tag:stsci.edu:asdf',
-     'http://asdf-standard.readthedocs.io/en/latest/generated/stsci.edu/asdf/'),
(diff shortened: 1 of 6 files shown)


jdavies-st:
Is there a way to show via tests that it was broken before and this now fixes it?

eslavich:
> Is there a way to show via tests that it was broken before and this now fixes it?

I think that's a good idea, we've had a lot of trouble with these.

jdavies-st (approved):
Looks good.  Tests fail on `master` but pass on this branch.  👍

PR #290, opened 2022-01-12
Add extension_metadata tag to older version maps/manifests

eslavich:
The extension_metadata tag was added in ASDF Standard 1.2.0:

https://github.com/asdf-format/asdf-standard/commit/c5313dda7a96d122012d594603518cea2a00194c#diff-314cd1da9cd8c76bcc041e47f574086723c322e98241f94c74e2563eea194cf5

but never actually included in the version_map-1.2.0 or version_map-1.3.0 file.  The manifests were created from the version_maps, so the oversight was propagated to them as well.


WilliamJamieson (approved):
LGTM

PR #292, opened 2022-02-08
Package asdf-standard

### LINKED ISSUE #293: Remove the use of `distutils`.
`distutils` is imported and used in two modules:

- [ ] [`tests/common.py`](https://github.com/asdf-format/asdf-standard/blob/d9f89ce8ae22c7e43bb45befd3c738459cf0c1a7/tests/common.py#L3)
- [ ] [`tests/test_version_map.py`](https://github.com/asdf-format/asdf-standard/blob/d9f89ce8ae22c7e43bb45befd3c738459cf0c1a7/tests/test_version_map.py#L3)

Since `distutils` is depreciated with removal set for python 3.12 (see https://docs.python.org/3/library/distutils.html), we should remove `distutils` soon.

WilliamJamieson:
Makes changes to repository to allow for packaging for pypi distribution.

Fixes #293.


docs/source/conf.py
```diff
@@ -8,15 +8,16 @@
...
-project = 'ASDF Standard'
+project = "ASDF Standard"
 
 import datetime
-copyright = f'{datetime.datetime.now().year}, Space Telescope Science Institute'
 
-author = 'Michael Droettboom, Erik Bray, et al'
+copyright = f"{datetime.datetime.now().year}, Space Telescope Science Institute"
+
+author = "Michael Droettboom, Erik Bray, et al"
```
eslavich, line 17: @nden @perrygreenfield should we change this to something like "The ASDF Developers"?

src/asdf_standard/integration.py
```diff
@@ -0,0 +1,32 @@
...
+        # In an editable install, the resources directory will exist off the
+        # repository root:
+        resources_root = Path(__file__).absolute().parent.parent.parent / "resources"
+        if not resources_root.is_dir():
+            raise RuntimeError("Missing resources directory")
+
+    return [
+        DirectoryResourceMapping(
+            resources_root / "schemas",
+            "asdf://{{ cookiecutter.uri_authority }}/{{ cookiecutter.uri_project }}/schemas/",
```
eslavich, line 26: We'll need to set these up with the correct URI prefixes

tests/common.py
```diff
@@ -1,19 +1,20 @@
 from pathlib import Path
 import re
-from distutils.version import StrictVersion
+from packaging.version import Version
 import yaml
 from urllib.parse import urljoin
```
eslavich, line 7: Not really related to this file, just need a place to put this comment:

We need a test module like https://github.com/asdf-format/asdf-transform-schemas/blob/master/tests/test_integration.py, that will catch issues with the DirectoryResourceMapping setup.

tox.ini
```diff
@@ -1,29 +1,40 @@
 [tox]
-skipsdist = True
-envlist = py39, py310, black, flake8, docs
+envlist = py38, py39, py310, black, flake8, docs
```
eslavich, line 2: This is mainly useful for developers who want to run `tox` without arguments and have it cover a reasonable subset of possible environments.  For that purpose it may not be necessary to test more than one Python version, but either way is good by me.

tests/common.py
WilliamJamieson: see 6f95534

src/asdf_standard/integration.py
WilliamJamieson: see 1600857

src/asdf_standard/integration.py
```diff
@@ -0,0 +1,29 @@
...
+        # repository root:
+        resources_root = Path(__file__).absolute().parent.parent.parent / "resources"
+        if not resources_root.is_dir():
+            raise RuntimeError("Missing resources directory")
+
+    return [
+        DirectoryResourceMapping(resources_root / "schemas" / "stsci.edu", "http://stsci.edu/schemas/", recursive=True),
+        DirectoryResourceMapping(
+            resources_root / "manifests" / "asdf-format.org" / "core",
+            "asdf://asdf-format.org/core/extensions/",
```
eslavich, line 27: The manifest URI prefix is actually `asdf://asdf-format.org/core/manifests/` but this is still passing because we have a resource plugin in asdf itself that is pulling from the submodule.

Any thoughts on this directory structure vs `asdf-format.org/manifests`?  That feels more natural to me since the authority is the root level name, and it will save us from having to add another DirectoryResourceMapping when we eventually have asdf-format.org schemas.  In any case we should pick a firm convention for that and make all the schemas repos match since like you pointed out we're not consistent right now.

WilliamJamieson: I like the `asdf-format.org/manifests`.

WilliamJamieson: See 48800a6 for the change to `core/manifests/`.

docs/source/conf.py
nden: That would be my preference.

WilliamJamieson: See 3b04a11

.github/workflows/ci.yml
```diff
@@ -41,6 +56,21 @@ jobs:
             os: ubuntu-latest
             toxenv: docs
 
+          - name: Bandit security audit
+            python-version: 3.9
+            os: ubuntu-latest
+            toxenv: bandit
```
nden, line 62: Bandit checks Python code only. I wonder if we should enable CodeQL for this repository.

WilliamJamieson: If so, then we should probably do it for asdf itself too.

eslavich (approved):
LGTM

PR #298, opened 2022-02-15
Fix readthedocs for asdf-standard

### LINKED ISSUE #297: RTD build is failing
See https://readthedocs.org/projects/asdf-standard/builds/16085697/ for example.  We need to update the install script that RTD uses to account for the new package setup.

Also we could use a CI check that runs the build on readthedocs.org so that we can catch this kind of issue in the future.

Thanks @pllim for the report :)

WilliamJamieson:
I believe that this will fix #297 by adding a `.readthedocs.yml` configuration file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,6 @@ jobs:
             os: ubuntu-latest
             toxenv: codestyle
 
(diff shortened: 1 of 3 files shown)


eslavich (approved):
🥇

PR #299, opened 2022-02-17
Fixed bug in manifest integration tests

WilliamJamieson:
Fixes a bug in the manifest integration tests.

It updates the integration tests to follow the convention in asdf-format/asdf-transform-schemas#37, so that tests are parameterized using the schema name rather than an arbitrary variable.


eslavich (approved):
LGTM

PR #302, opened 2022-02-23
Update toolchains

WilliamJamieson:
Brings asdf-standard tool configurations in line with asdf, PR: asdf-format/asdf#1091.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,28 +11,24 @@ repos:
   - id: end-of-file-fixer
   - id: trailing-whitespace
 
-- repo: https://github.com/psf/black
-  rev: 22.1.0
-  hooks:
-    - id: black
-      exclude: "reference_files"
-
 - repo: https://github.com/pycqa/isort
   rev: 5.10.1
   hooks:
     - id: isort
-      args: ["--profile", "black", "--filter-files"]
-      exclude: "reference_files"
+
+- repo: https://github.com/psf/black
+  rev: 22.1.0
+  hooks:
+    - id: black
 
 - repo: https://gitlab.com/pycqa/flake8
   rev: 4.0.1
   hooks:
     - id: flake8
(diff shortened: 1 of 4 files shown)


bandit.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+exclude_dirs:
+  - .eggs
+  - .git
+  - .pytest_cache
+  - .tox
+  - reference_files
+  - tests
+  - build
+  - compatibility_tests
```
eslavich, line 11: This directory doesn't exist for asdf-standard

pyproject.toml
```diff
@@ -7,3 +7,21 @@ write_to = "src/asdf_standard/_version.py"
...
+force-exclude = '''
+^/(
+  (
+      \.eggs
+    | \.git
+    | \.pytest_cache
+    | \.tox
+    | reference_files
+    | build
+    | dist
```
eslavich, line 19: Should we be excluding the resources directory here?

.pre-commit-config.yaml
```diff
@@ -11,28 +11,24 @@ repos:
...
+  rev: 22.1.0
+  hooks:
+    - id: black
 
 - repo: https://gitlab.com/pycqa/flake8
   rev: 4.0.1
   hooks:
     - id: flake8
-      args: ["--count", "--ignore", "E501, E203, W503"]
       exclude: "reference_files"
```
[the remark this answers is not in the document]
WilliamJamieson: There should not be any python files in resources, so flake8 should not do anything there. The configuration is in the setup.cfg as it was before.

bandit.yaml
WilliamJamieson: Its just yaml files so it should be ignored.

pyproject.toml
WilliamJamieson: Again, resources is just yaml files so it will be ignored anyways.

PR #305, opened 2022-02-23
Check in CRLF --> LF diff to logo.png

eslavich:


WilliamJamieson:
In addition to this change should we add `*.png binary` to the `.gitattributes`?

eslavich:
Others suggested a solution that does not corrupt the PNG file :)

PR #321, opened 2022-07-07
Move packaging into pyproject.toml

WilliamJamieson:
Moves everything it can from setup.cfg to pyproject.toml


zacharyburnett (approved):
if the tests pass then I think this looks good

EDIT: could this affect downstream usage?

nden (approved):

nstarman (commented):
Great to see. I have a non-pushed PR to do this for astropy! but not alll the ``tool.pytest.ini_options`` stuff appears to be compatible.

PR #323, opened 2022-07-26
Fix schema for compatibility with rad

WilliamJamieson:
Changes to the rad schemas, require the dropping of the `tag` entry in the `time-1.1.0` schema. This should not cause any issues as the `tag` in the schema is unnecessary for proper schema functioning, but does prevent the use of the schema to build another tagged object.


nden (approved):

perrygreenfield (approved):
LGTM

PR #326, opened 2022-08-08
Remove the rest of the `tag` references.

WilliamJamieson:
Fixes #323. Note `fits` still has tag because this is required for compatibility with fits.


perrygreenfield (approved):
Looks very straightforward.

Issue #345, opened 2022-11-10
Overly permissive ndarray schema leads to improper time validation


braingram:
The ndarray schema accepts any object that does not have a source key (and some that do). For example:
```python
import asdf
import asdf.tests


bad_string = """
!core/ndarray-1.0.0
 foo: 1
 bar: 2
"""

buff = asdf.tests.helpers.yaml_to_asdf(f"example: {bad_string.strip()}")

# confirm that validate on read is True
c = asdf.get_config()
c.validate_on_read = True

with asdf.AsdfFile() as af:
    try:
        af._open_impl(af, buff, mode='rw')
    except asdf.ValidationError as e:
        print("File appropriately failed to validate")
```
Does not produce a validation error and instead produces an exception during conversion to an array:
```
  File "/Users/bgraham/projects/221108_stdatamodel_bugs/asdf/asdf/tags/core/ndarray.py", line 395, in from_tree
    byteorder = node["byteorder"]
KeyError: 'byteorder
```

This schema is referenced in the time schema as part of an anyOf composition:
https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/time/time-1.1.0.yaml#L127
which means that the subsequent object schema defined in time is not used and invalid yaml is produced for time objects that violate the time schema. For example, serializing
```python
astropy.time.Time(730120.0003703703, format="plot_date")
```
produces

```yaml
t: !time/time-1.1.0 {base_format: plot_date, value: '2000-01-01T00:00:32.000'}
```
which contains a base_format that is [not listed in the time schema](https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml#L180) (this example was taken from the time [tests in asdf-astropy](https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/converters/time/tests/test_time.py#L37) some of which are passing because of this bug).

PR #349, opened 2022-11-17
Fix the `base_format` list.

### LINKED ISSUE #346: time schema is inconsistent with asdf-astropy
This has been hidden by and is related to https://github.com/asdf-format/asdf-standard/issues/345

asdf-astropy serializes astropy.time.Time values in a way that is inconsistent with the time schema. One example is that the schema lists [allowed base_formats](https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml#L180):
```yaml
    enum:
      - iso
      - yday
      - byear
      - jyear
      - decimalyear
      - jd
      - mjd
      - gps
      - unix
      - cxcsec
```
which are incomplete compared to the base_format values [produced by asdf-astropy](https://github.com/astropy/asdf-astropy/blob/main/asdf_astropy/converters/time/time.py#L22) (which can contain any astropy.time.Time format). For example, serializing
```python
astropy.time.Time(730120.0003703703, format="plot_date")
```
produces

```yaml
t: !time/time-1.1.0 {base_format: plot_date, value: '2000-01-01T00:00:32.000'}
```

Given the assumed intended behavior of base_format (to store the astropy.time.Time format) this looks more like a schema bug than an issue with the asdf-astropy implementation.

The base_format is one (but possibly not the only) error with the time schema that was hidden by the ndarray schema issue.

### LINKED ISSUE #352: Time schema validates arbitrary ndarray objects that converters cannot interpret as time
Neither the converter in astropy or asdf-astropy supports the reference to the ndarray object, I vote for removing the reference:
https://github.com/asdf-format/asdf-standard/blob/ecdf2648b75fc99053fa33a5ff06cab1561af5eb/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml#L197
For example, the following is valid yaml if the ndarray reference is kept:
```yaml
!time/time-1.1.0
  data: [1, 2, 3]
```
So this validates during reading, but fails during conversion to astropy.Time.

WilliamJamieson:
Fixes #346.
Fixes #352 per comment https://github.com/asdf-format/asdf-standard/issues/352#issuecomment-1323770906

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ The in progress ASDF Standard is v1.6.0
 The stable ASDF Standard is v1.5.0
 
 - Add new ``quantity-1.2.0`` schema to support ``datatype`` option. [#351]
-
+- Bugfix for ``base_format`` in ``time-1.2.0`` schema. [#349]
 
 1.0.3 (2022-08-08)
 ------------------
--- a/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
+++ b/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
@@ -210,8 +210,6 @@ anyOf:
 
   - $ref: "#/definitions/array_of_strings"
 
-  - $ref: "../core/ndarray-1.0.0#/anyOf/1"
-
   - type: object
     properties:
       value:
--- a/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
+++ b/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
@@ -187,6 +187,7 @@ definitions:
       - jyear
       - mjd
       - unix
+      - unix_tai
       - yday
 
   other_format:
@@ -201,9 +202,8 @@ definitions:
       - isot
       - jyear_str
       - plot_date
-      - unix_tai
       - ymdhms
-      - datetime-64
+      - datetime64
 
 anyOf:
   - $ref: "#/definitions/string_formats"
--- a/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
+++ b/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
@@ -178,16 +178,32 @@ definitions:
         do for the purposes of ASDF?]
 
     enum:
-      - iso
-      - yday
       - byear
-      - jyear
-      - decimalyear
-      - jd
-      - mjd
-      - gps
-      - unix
       - cxcsec
+      - decimalyear
+      - gps
+      - iso
+      - jd
+      - jyear
+      - mjd
+      - unix
+      - yday
+
+  other_format:
+    description: |
+      The other formats supported by astropy.time:
+      https://docs.astropy.org/en/latest/time/index.html#time-format
+
+    enum:
+      - byear_str
+      - datetime
+      - fits
+      - isot
+      - jyear_str
+      - plot_date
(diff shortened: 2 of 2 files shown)


braingram:
@perrygreenfield  Should this be an new time schema version (1.3)?

Also, @eslavich should we remove the ref to ndarray? If we leave it in the object definition changed here will never be used for validation because of issue https://github.com/asdf-format/asdf-standard/issues/345

WilliamJamieson:
> @perrygreenfield Should this be an new time schema version (1.3)?
> 
> Also, @eslavich should we remove the ref to ndarray? If we leave it in the object definition changed here will never be used for validation because of issue #345

This should be time-1.2.0 as these are simply corrections to the changes that brought `1.2.0` into existence

braingram:
Perhaps this is a separate issue but time is referenced in wcs (one example [here](https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/wcs/frame-1.1.0.yaml#L98) I'm only seeing 2 refs). The refs are currently pointing to time-1.1. Should these be updated to time-1.2?

WilliamJamieson:
> Perhaps this is a separate issue but time is referenced in wcs (one example [here](https://github.com/asdf-format/asdf-standard/blob/master/resources/schemas/stsci.edu/asdf/wcs/frame-1.1.0.yaml?rgh-link-date=2022-11-17T20%3A53%3A26Z#L98) I'm only seeing 2 refs). The refs are currently pointing to time-1.1. Should these be updated to time-1.2?

Since `additionalProperties` is not set and its default is `true` then the `base_format` will field will not cause any issues in wcs. When asdf-standard 1.6 is finalized and becomes the stable version then we should update things.

braingram:
Given that neither the converter in astropy or asdf-astropy supports the reference to the ndarray object, I vote for removing the reference:
https://github.com/asdf-format/asdf-standard/blob/ecdf2648b75fc99053fa33a5ff06cab1561af5eb/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml#L197
For example, the following is valid if the ndarray reference is kept:
```yaml
!time/time-1.1.0
  data: [1, 2, 3]
```
However this fails to convert to a time.

WilliamJamieson:
> Given that neither the converter in astropy or asdf-astropy supports the reference to the ndarray object, I vote for removing the reference:
> 
> https://github.com/asdf-format/asdf-standard/blob/ecdf2648b75fc99053fa33a5ff06cab1561af5eb/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml#L197

This discussion is outside the scope of this PR. It is being removed in #350 and therefore should be discussed there.

braingram:
I think the ndarray ref is relevant to this issue as it's an inconsistency between the schema and converters. https://github.com/asdf-format/asdf-standard/issues/346

This PR fixes the base_format errors but not 346 as the schema and converters are still inconsistent.

I'd be ok with merging this PR and reopening the issue.

WilliamJamieson:
> This PR fixes the base_format errors but not 346 as the schema and converters are still inconsistent.

Please elaborate, what inconsistencies still exist?

braingram (commented):
I see your point that separating the ndarray ref will make this easier to sort out. This looks good to me and I will open a new issue for the ndarray ref.

braingram (approved):
I'm not sure my previous comment was marked with approval. This looks great to me.

PR #350, opened 2022-11-21
Update ndarray schema to 1.1 to fix issue 345

### LINKED ISSUE #345: Overly permissive ndarray schema leads to improper time validation

braingram:
Fixes #345 

~~This PR should be merged after #349 so the time schema errors fixed in that PR do not result in test failures once this PR fixes #345.~~

### DIFF
--- a/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
+++ b/resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
@@ -71,7 +71,7 @@ examples:
     - asdf-standard-1.6.0
     - |
         !time/time-1.2.0
-          value: !core/ndarray-1.0.0
+          value: !core/ndarray-1.1.0
             data: [2000, 2001]
             datatype: float64
           format: jyear
@@ -87,13 +87,13 @@ examples:
           location:
             x: !unit/quantity-1.2.0
               value: 6378100
-              unit: !unit/unit-1.0.0 m
+              unit: !unit/unit-1.1.0 m
             y: !unit/quantity-1.2.0
               value: 0
-              unit: !unit/unit-1.0.0 m
+              unit: !unit/unit-1.1.0 m
             z: !unit/quantity-1.2.0
               value: 0
-              unit: !unit/unit-1.0.0 m
+              unit: !unit/unit-1.1.0 m
 
 definitions:
   iso_time:
@@ -219,7 +219,7 @@ anyOf:
         anyOf:
           - $ref: "#/definitions/string_formats"
           - $ref: "#/definitions/array_of_strings"
-          - $ref: "../core/ndarray-1.0.0"
+          - $ref: "../core/ndarray-1.1.0"
           - type: number
 
       format:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ The stable ASDF Standard is v1.5.0
 
 - Add new ``quantity-1.2.0`` schema to support ``datatype`` option. [#351]
 - Bugfix for ``base_format`` in ``time-1.2.0`` schema. [#349]
+- Add new ``ndarray-1.1.0`` schema to fix #345 [#350]
 
 1.0.3 (2022-08-08)
 ------------------
--- a/docs/source/schemas/core.rst
+++ b/docs/source/schemas/core.rst
@@ -10,13 +10,13 @@ asdf library.
 
    core/asdf-1.1.0
    core/complex-1.0.0
-   core/ndarray-1.0.0
-   core/table-1.0.0
-   core/column-1.0.0
+   core/ndarray-1.1.0
+   core/table-1.1.0
+   core/column-1.1.0
    core/constant-1.0.0
    core/software-1.0.0
    core/history_entry-1.0.0
    core/extension_metadata-1.0.0
-   core/integer-1.0.0
+   core/integer-1.1.0
    core/externalarray-1.0.0
    core/subclass_metadata-1.0.0
--- a/docs/source/schemas/fits.rst
+++ b/docs/source/schemas/fits.rst
@@ -8,4 +8,4 @@ FITS. It can safely be ignored by most ASDF implementations.
 
 .. asdf-autoschemas::
 
-   fits/fits-1.0.0
+   fits/fits-1.1.0
--- a/docs/source/schemas/legacy.rst
+++ b/docs/source/schemas/legacy.rst
@@ -8,6 +8,13 @@ The following legacy schemas and not part of the most recent ASDF standard versi
 .. asdf-autoschemas::
 
     core/asdf-1.0.0
+    core/column-1.0.0
+    core/integer-1.0.0
+    core/ndarray-1.0.0
+    core/table-1.0.0
+    fits/fits-1.0.0
     time/time-1.0.0
     time/time-1.1.0
+    unit/defunit-1.0.0
     unit/quantity-1.1.0
+    unit/unit-1.0.0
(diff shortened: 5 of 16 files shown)


WilliamJamieson:
Can you separate the change adding the new `ndarray` schema into a separate commit followed by all the supporting changes to make the history a bit easier to parse.

Also, can you post the diff between `ndarray-1.0.0` and `ndarray-1.1.0`?

resources/schemas/stsci.edu/asdf/time/time-1.2.0.yaml
```diff
@@ -178,24 +178,38 @@ definitions:
...
+      - ymdhms
+      - datetime64
 
 anyOf:
   - $ref: "#/definitions/string_formats"
 
   - $ref: "#/definitions/array_of_strings"
 
-  - $ref: "../core/ndarray-1.0.0#/anyOf/1"
-
```
WilliamJamieson, line 198: Isn't this PR supposed to make it so that the removal of this becomes unnecessary?

braingram:
I may be missing something but I am unable to (locally) get the asdf-development tests to pass without updating the legacy converters to support the new schemas. Specifically:
- in astropy.io.misc.asdf)
  - table
  - column
- in asdf
  - integer
  - ndarray
 
 The failing test is `test_version_map_core_support[version6]` which fails with a rather unhelpful error:
`AttributeError: 'MockContext' object has no attribute '_warn_tag_mismatch'`
The MockContext [here](https://github.com/asdf-format/asdf/blob/master/asdf/tests/test_versioning.py#L279), meant to mimic an AsdfFile does not have that method. If a method is added (that does nothing, matching the default behavior in AsdfFile), the error becomes a more helpful:
```
E           AssertionError: ASDF Standard version 1.6.0 requires support for tag:stsci.edu:asdf/core/integer-1.1.0, but no ExtensionType exists that explicitly supports that version.
E           assert 'tag:stsci.ed...integer-1.0.0' == 'tag:stsci.ed...integer-1.1.0'
E             - tag:stsci.edu:asdf/core/integer-1.1.0
E             ?                                   ^
E             + tag:stsci.edu:asdf/core/integer-1.0.0
E             ?                                   ^
```
If I then add a `supported_versions` class attribute to the [IntegerType in asdf](https://github.com/asdf-format/asdf/blob/master/asdf/tags/core/integer.py#L46) the test no longer fails for the integer tag.

However, the NDArrayType class does not appear to support a similar fix. If I add a supported_versions class attribute I get an even less helpful error at time of class definition:
```python
TypeError: __class__ set to <class 'asdf.tags.core.ndarray.NDArrayType'> defining 'NDArrayType' as <class 'asdf.tags.core.ndarray.NDArrayType'>
```

To move this PR forward, am I correct that the asdf-development tests will fail until the converters are updated or am I missing something?

Also, thanks for the feedback on the history changes, I'll make them once the content of the PR is sorted out.

Aside from the examples in the description the diff for ndarray-1.0.0 and ndarray-1.1.0 is as follows:
```diff
>     oneOf:
>       - required: [source]
>       - required: [data]
>
```

WilliamJamieson:
> I may be missing something but I am unable to (locally) get the asdf-development tests to pass without updating the legacy converters to support the new schemas. Specifically:
> 
> * in astropy.io.misc.asdf)
>   
>   * table
>   * column
> * in asdf
>   
>   * integer
>   * ndarray
> 
> The failing test is `test_version_map_core_support[version6]` which fails with a rather unhelpful error: `AttributeError: 'MockContext' object has no attribute '_warn_tag_mismatch'` The MockContext [here](https://github.com/asdf-format/asdf/blob/master/asdf/tests/test_versioning.py?rgh-link-date=2022-11-21T20%3A18%3A26Z#L279), meant to mimic an AsdfFile does not have that method. If a method is added (that does nothing, matching the default behavior in AsdfFile), the error becomes a more helpful:
> 
> ```
> E           AssertionError: ASDF Standard version 1.6.0 requires support for tag:stsci.edu:asdf/core/integer-1.1.0, but no ExtensionType exists that explicitly supports that version.
> E           assert 'tag:stsci.ed...integer-1.0.0' == 'tag:stsci.ed...integer-1.1.0'
> E             - tag:stsci.edu:asdf/core/integer-1.1.0
> E             ?                                   ^
> E             + tag:stsci.edu:asdf/core/integer-1.0.0
> E             ?                                   ^
> ```
> 
> If I then add a `supported_versions` class attribute to the [IntegerType in asdf](https://github.com/asdf-format/asdf/blob/master/asdf/tags/core/integer.py?rgh-link-date=2022-11-21T20%3A18%3A26Z#L46) the test no longer fails for the integer tag.
> 
> However, the NDArrayType class does not appear to support a similar fix. If I add a supported_versions class attribute I get an even less helpful error at time of class definition:
> 
> ```python
> TypeError: __class__ set to <class 'asdf.tags.core.ndarray.NDArrayType'> defining 'NDArrayType' as <class 'asdf.tags.core.ndarray.NDArrayType'>
> ```
> 
> To move this PR forward, am I correct that the asdf-development tests will fail until the converters are updated or am I missing something?
> 

This should be paired with an `asdf` PR. Failures in `astropy.io.misc.asdf` are irrelevant.

braingram:
Here is a partial diff between ndarray-1.0.0 and ndarray-1.1.0 (most differences are updating references/tags from 1.0 to 1.1):
```diff
374a386,389
>     oneOf:
>       - required: [source]
>       - required: [data]
>
```

braingram:
There is 1 remaining reference to ndarray-1.0 in the asdf-schema-1.0.0 metaschema:
https://github.com/asdf-format/asdf-standard/blob/1e223562401c64ca1b180356779c2082f0195310/resources/schemas/stsci.edu/asdf/asdf-schema-1.0.0.yaml#L39
This references a portion of the ndarray-1.0.0 schema that is not affected by #345.

The CI failure of asdf-development tests is unavoidable as asdf does not yet have support for the new schemas (support will be added in this PR: https://github.com/asdf-format/asdf/pull/1250).

I think it makes sense to merge this prior to the asdf PR so that the CI for that PR can hopefully have fewer errors (it will still fail some tests unless https://github.com/asdf-format/asdf/pull/1247 is merged to allow xfailing certain new tags that are not supported in asdf).

WilliamJamieson (approved):

PR #359, opened 2022-12-13
Add more pre-commit checks

WilliamJamieson:
Switch from `flake8` to `ruff` and enable more pre-commit checks


braingram (approved):
Looks good to me.

PR #370, opened 2023-03-13
use default branches in ci jobs

braingram:
This will make fixing https://github.com/asdf-format/asdf/issues/1478 easier.


braingram:
weldx failures are unrelated and might be related to https://github.com/asdf-format/asdf/pull/1340

WilliamJamieson (approved):

PR #409, opened 2023-11-10
update reference files for new asdf standard versions

braingram:
Fixes a bug in the reference file generator where generated files used the latest asdf standard.

Generates and adds reference files for asdf standard versions 1.1.0 through 1.5.0 (inclusive).

I've added a few comments describing changes to the reference files (and the script to generate them).


reference_files/1.0.0/complex.yaml
```diff
@@ -3,190 +3,190 @@
...
-asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
-  name: asdf, version: 1.0.0}
+asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
+  name: asdf, version: 3.0.1}
 datatype<c16: !core/ndarray-1.0.0
   data: [!core/complex-1.0.0 0j, !core/complex-1.0.0 0j, !core/complex-1.0.0 (nan+nanj),
-    !core/complex-1.0.0 (nan+infj), !core/complex-1.0.0 (nan-infj), !core/complex-1.0.0 -1.79769313486e+308j,
-    !core/complex-1.0.0 1.79769313486e+308j, !core/complex-1.0.0 2.22044604925e-16j,
-    !core/complex-1.0.0 1.11022302463e-16j, !core/complex-1.0.0 2.22507385851e-308j,
+    !core/complex-1.0.0 (nan+infj), !core/complex-1.0.0 (nan-infj), !core/complex-1.0.0 -1.7976931348623157e+308j,
```
braingram, line 10: The values in the existing `complex.yaml` are not correct for a 64 bit float (real or imaginary portion of a complex 128, yaml will treat all numbers as 64 bit so the expanded representation should be seen for all  bit-depths of complex values in this file).
```python
>> np.finfo(np.float64).max == 1.79769313486e+308
False
>> np.finfo(np.float64).max == 1.7976931348623157e+308
True
```

reference_files/1.0.0/unicode_spp.yaml
```diff
@@ -3,14 +3,14 @@
...
 %TAG ! tag:stsci.edu:asdf/
 --- !core/asdf-1.0.0
-asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
-  name: asdf, version: 1.0.0}
+asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
+  name: asdf, version: 3.0.1}
 datatype<U: !core/ndarray-1.0.0
   data: ['', "\U00010020"]
-  datatype: [ucs4, 2]
+  datatype: [ucs4, 1]
```
braingram, line 10: It is unclear to me what `unicode_spp` is supposed to be illustrating. My closest guess is that the character `𐀠` is [U+10020](https://unicode-explorer.com/c/10020) is part of plane 1 the [Supplementary Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Supplementary_Multilingual_Plane) so perhaps this was suppored to be `smp` not `spp`. On my system this is stored in 4 byte unicode which explains the `[ucs4, 1]` change in this file. I'm not sure what system or version of numpy was using 8 bytes to store this character.
```python
>> np.unicode_('𐀠').dtype.itemsize
4
```
The `generate.py` file could be updated to use `U2` instead of `U` to generate a file that uses 8 bytes:
```python
>> np.array(['𐀠'], dtype='U2').dtype.itemsize
8
```
At this point I suggest we hold off making this change to `generate.py` and instead include the updated file in this PR as it offers a slightly different example for unicode in ASDF compared to the `bmp` file (which has a combined character which results in `[ucs4, 2]`).

reference_files/generate/generate.py
```diff
@@ -155,8 +155,8 @@ def generate(version):
         filename = os.path.join(outdir, name)
         func(filename + ".asdf")
         with asdf.open(filename + ".asdf") as af:
-            af.resolve_and_inline()
-            af.write_to(filename + ".yaml")
+            af.resolve_references()
```
braingram, line 158: This replaces the only use of `resolve_and_inline` I'm able to find in any package outside of asdf. See https://github.com/asdf-format/asdf/pull/1690 for a description of the `resolve_and_inline` deprecation which includes the suggestion to update code similar to the update in this PR (replace `resolve_and_inline` with `resolve_references` and `all_array_storage='inline'`).

perrygreenfield (approved):
LGTM

PR #415, opened 2023-12-12
Freeze 1.6.0, introduce 1.7.0 (as dev)

braingram:
EDIT: this PR was too ambitious and has been simplified (removing the semantic versioning change) and now includes only the freezing of 1.6.0 and introduction of 1.7.0 (as the new development version).

I am leaving this as draft because the release of 1.6.0 will impact a large number of packages. A test PR is open for asdf where the downstream impact is being tested: https://github.com/asdf-format/asdf/pull/1744


braingram:
@eslavich I dismissed your review here to not give the impression that the current state of this PR is approved. The current PR is quite different from the one you reviewed for a few reasons:
- the change from non-semantic to semantic versioning should have been made in a new manifest extension schema
- the release of 1.6.0 is causing a lot of package failures which should be reviewed before this PR can be re-reviewed

Thanks for giving this a look and I believe the above changes are still work doing (especially the new manifest extension schema). I just think they will require more extensive changes that will span multiple PRs.

braingram:
Superseded by: https://github.com/asdf-format/asdf-standard/pull/422

PR #443, opened 2024-09-26
remove statement discouraging anchors and aliases

braingram:
As was brought up in https://github.com/asdf-format/asdf/issues/1840
the standard currently discourages anchors and aliases. This PR removes the statement discouraging their use.


perrygreenfield (approved):
LGTM

PR #480, opened 2025-08-26
add galexsec, reorganize time formats

braingram:


braingram:
> I found some time...but I forget which standard...to review this.
> 
> LGTM

pgftime™

Issue #484, opened 2025-08-27
Add missing transform schemas.


braingram:
asdf-1.0.0 refs wcs-1.0.0:
https://github.com/asdf-format/asdf-standard/blob/6d2a44ea11f6e0bee35525517b48621ce1076770/resources/schemas/stsci.edu/asdf/core/asdf-1.0.0.yaml#L46
which refs step-1.0.0:
https://github.com/asdf-format/asdf-standard/blob/6d2a44ea11f6e0bee35525517b48621ce1076770/resources/schemas/stsci.edu/asdf/wcs/wcs-1.0.0.yaml#L28
which refs transform-1.0.0:
https://github.com/asdf-format/asdf-standard/blob/6d2a44ea11f6e0bee35525517b48621ce1076770/resources/schemas/stsci.edu/asdf/wcs/step-1.0.0.yaml#L27
which is not part of the asdf-standard package.

This resulted in asdf adding asdf-transform-schemas as a dependency:
https://github.com/asdf-format/asdf/blob/a346a6b027873448208d181dab89355279ec5fe2/pyproject.toml#L18
but could instead by fixed by adding copies of transform-1.0.0
https://github.com/asdf-format/asdf-transform-schemas/blob/main/resources/stsci.edu/schemas/transform-1.0.0.yaml
and the referenced domain-1.0.0
https://github.com/asdf-format/asdf-transform-schemas/blob/main/resources/stsci.edu/schemas/domain-1.0.0.yaml
schemas. This would allow:
- dropping asdf-transform-schemas as a dependency in asdf
- likely unskipping some tests here https://github.com/asdf-format/asdf-standard/blob/6d2a44ea11f6e0bee35525517b48621ce1076770/pyproject.toml#L69

As these are released schemas that should never change having copies is not an issue.

PR #485, opened 2025-08-27
Add missing transform schemas

braingram:
Closes https://github.com/asdf-format/asdf-standard/issues/484

Add copies of transform-1.0.0 and domain-1.0.0 to the schemas. These are referenced by the core schema asdf-1.0.0 but were previously moved to asdf-transform-schemas. This resulted in a need to add asdf-transform-schemas to a dependency of asdf (but left this package incomplete).

See https://github.com/asdf-format/asdf/pull/1965 for a test showing that this fixes failures that would otherwise be seen if asdf drop asdf-transform-schemas as a dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,22 +62,9 @@ where = ["src"]
 
 [tool.pytest.ini_options]
 asdf_schema_root = 'resources/schemas'
-asdf_schema_skip_tests = """
-    stsci.edu/asdf/asdf-schema-1.0.0.yaml
-    stsci.edu/asdf/wcs/celestial_frame-1.0.0.yaml
-    stsci.edu/asdf/wcs/celestial_frame-1.1.0.yaml
-    stsci.edu/asdf/wcs/frame-1.0.0.yaml
-    stsci.edu/asdf/wcs/frame-1.1.0.yaml
-    stsci.edu/asdf/wcs/spectral_frame-1.1.0.yaml
-    stsci.edu/asdf/wcs/step-1.1.0.yaml
-    stsci.edu/asdf/wcs/step-1.2.0.yaml
-    stsci.edu/asdf/wcs/wcs-1.1.0.yaml
-    stsci.edu/asdf/wcs/wcs-1.2.0.yaml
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -62,22 +62,15 @@ where = ["src"]
...
 asdf_schema_root = 'resources/schemas'
+asdf_schema_tests_enabled = 'true'
 asdf_schema_skip_tests = """
-    stsci.edu/asdf/asdf-schema-1.0.0.yaml
-    stsci.edu/asdf/wcs/celestial_frame-1.0.0.yaml
-    stsci.edu/asdf/wcs/celestial_frame-1.1.0.yaml
-    stsci.edu/asdf/wcs/frame-1.0.0.yaml
-    stsci.edu/asdf/wcs/frame-1.1.0.yaml
-    stsci.edu/asdf/wcs/spectral_frame-1.1.0.yaml
     stsci.edu/asdf/wcs/step-1.1.0.yaml
```
braingram, line 67: There are a few remaining skips for schemas that aren't referenced by any core schemas. However the changes in this PR allow removal of more than half the skips.

pyproject.toml
```diff
@@ -62,22 +62,15 @@ where = ["src"]
...
     stsci.edu/asdf/wcs/step-1.2.0.yaml
     stsci.edu/asdf/wcs/wcs-1.1.0.yaml
     stsci.edu/asdf/wcs/wcs-1.2.0.yaml
-    stsci.edu/yaml-schema/draft-01.yaml
 """
-asdf_schema_tests_enabled = 'true'
 asdf_schema_ignore_unrecognized_tag = 'true'
-addopts = '--color=yes'
+addopts = '--color=yes -r a'
```
braingram, line 73: This just prints out extra information about skipped, failed or errored tests.

perrygreenfield:
Would it be useful to add comments in both the "legacy" schema to be explicit that these are only for legacy cases and should not be used in future cases?

perrygreenfield:
That seems reasonable to be consistent with the others.

perrygreenfield (approved):
LGTM

PR #489, opened 2025-09-15
Update pre commit

braingram:
This updates the pre-commit hooks to more closely align with those in asdf.

Replaces pyupgrade, isort and bandit with equivalent ruff rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,7 @@ select = [
     "RUF",  # ruff specific, includes yesqa
 ]
 extend-ignore = [
-    "RUF005",
+    "RUF005", # Checks for uses of the + operator to concatenate collections.
 ]
 
 [tool.ruff.lint.per-file-ignores]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -92,13 +92,26 @@ force-exclude = '''
...
+    "RUF",  # ruff specific, includes yesqa
+]
+extend-ignore = [
+    "RUF005", # Checks for uses of the + operator to concatenate collections.
+]
+
+[tool.ruff.lint.per-file-ignores]
+"test_*.py" = ["S101"]
+"tests/*" = ["S101"]
+"scripts/*" = ["S603"]
```
zacharyburnett, line 114:
```suggestion
"test_*.py" = [
    "S101", # Checks for uses of the assert keyword.
]
"tests/*" = [
    "S101", # Checks for uses of the assert keyword.
]
"scripts/*" = [
    "S603", # Check for method calls that initiate a subprocess without a shell.
]
```

PR #503, opened 2026-03-30
Update RTD build OS from Ubuntu 20.04 to 24.04

sydduckworth:
## Changelog
* Updated RTD build OS from Ubuntu 20.04 to 24.04
* Updated `docs/conf.py` to use `importlib_metadata` instead of `pkg_resources`


braingram (approved):
Thanks!

